PR #3, opened 2025-10-17
Multiple sources for same name,version tuple...

dzbarsky:


csmoe:
Just for a note, git dep can be specified with branch `bitlfags = { git = "https://github.com/bitflags/bitflags", branch = "cargo/2.10.0" }`
which will results a dep in lock as: `bitflags-2.10.0 (git+https://github.com/bitflags/bitflags?branch=cargo/2.10.0)`, then the `#` parsing will miss this case.

dzbarsky:
> Just for a note, git dep can be specified with branch `bitlfags = { git = "https://github.com/bitflags/bitflags", branch = "cargo/2.10.0" }` which will results a dep in lock as: `bitflags-2.10.0 (git+https://github.com/bitflags/bitflags?branch=cargo/2.10.0)`, then the `#` parsing will miss this case.

If there's no commit hash in the lockfile, how would the build system know when to refetch the repo? It sounds like a recipe for over invalidation and/or non-reproducible builds

csmoe:
`bitflags-2.10.0 (git+https://github.com/bitflags/bitflags?branch=cargo/2.10.0)` appears in other package's deps. bitflags's package name has a hash indeed

dzbarsky:
Ah gotcha! Ok I can take a look when I have a bit of time. Alternately, are you interested in sending a PR?

csmoe:
Glad to! I'm trying to make my local patch working, will contribute if succeed.

PR #62, opened 2026-03-19
Prevent adding @crates// dependency entries for dependencies in the workspace

walter-zeromatter:
This comes up if you, for example, have a vendored dependency which you've modified to point at something in your workspace (like I do)


dzbarsky:
Fixed in https://github.com/hermeticbuild/rules_rs/commit/c2ee3f310838fc11c0f5e7522f8c498f42d67470

PR #81, opened 2026-03-31
Add support for git and sparse purl qualifiers

darkrift:
This implements the get_purl function from the [cyclonedx-rust-cargo](https://github.com/CycloneDX/cyclonedx-rust-cargo/blob/e58bd5590212f82c5b7e16dd3e2e819b0dbea5b1/cargo-cyclonedx/src/purl.rs#L8) reference project relevant to `sparse` and `git` source url

It vendors the purl library from [supply-chain](https://github.com/bazel-contrib/supply-chain/tree/main/metadata/purl) because that library visibility is not allowing us to reuse it for now


rs/private/crate_git_repository.bzl
```diff
@@ -15,6 +15,50 @@ _INHERITABLE_FIELDS = [
...
+    if not url.startswith("git+"):
+        fail("expected source.repr to start with 'git+': %r" % url)
+
+    q = url.find("?")  # -1 if not found
+    f = url.find("#")  # -1 if not found
+
+    has_q = (q != -1)
+    has_f = (f != -1)
+
+    # Has both query params and commit hash: strip query, keep commit as @
```
dzbarsky: just out of curiousity...if you do `cargo metadata` on your testcase, are there no other fields that can help retrieve this commit so we don't have to do this parse?

PR #104, opened 2026-04-27
chore: open BCR prs in draft mode

kormide:
The BCR has a mechanism where a module maintainer can mark the PR as "ready for review" as a way to auto-approve and merge it. This is the recommended approach when the PRs are opened as an individual rather than a machine account as in https://github.com/bazelbuild/bazel-central-registry/pull/8569.

See https://github.com/bazel-contrib/publish-to-bcr#4-consider-whether-to-open-the-pr-as-a-draft.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -21,12 +21,12 @@ on:
         type: string
 jobs:
   publish:
-    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.0.0
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.2.0
     with:
       tag_name: ${{ inputs.tag_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -21,12 +21,12 @@ on:
         type: string
 jobs:
   publish:
-    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.0.0
+    uses: bazel-contrib/publish-to-bcr/.github/workflows/publish.yaml@v1.2.0
```
kormide, line 24: `v1.2.0` supports immutable GitHub releases. If you want to enable that, see https://github.com/bazel-contrib/publish-to-bcr#immutable-releases.

kormide:
fyi @dzbarsky

dzbarsky:
thanks! I feel like the BCR publish mechanism definitely changed (even after the BCR app). Endless treadmill...

PR #108, opened 2026-04-28
Add wasm bindgen extension

isaacparker0:


dzbarsky:
Fixed on main, thanks for the template!

PR #114, opened 2026-05-01
crate_git_repository: recurse submodules to match Cargo

walter-zeromatter:
Cargo always recursively checks out submodules for git dependencies, so a crate whose source uses submodules (e.g. khronos_api in rust-windowing/gl-rs, which vendors KhronosGroup/WebGL via submodule for the WebGL extensions registry) builds cleanly under Cargo but produces empty source trees here. The previous implementation only ran `git worktree add`, which does not populate submodules.

Run `git submodule update --init --recursive` after the worktree is created when a `.gitmodules` file is present. To preserve repository reproducibility, also remove the per-submodule `.git` gitlink files (which contain machine-specific absolute paths into the cache repo's modules dir) at the end of the rule, alongside the existing top-level `.git` removal.

### DIFF
--- a/rs/private/crate_git_repository.bzl
+++ b/rs/private/crate_git_repository.bzl
@@ -82,6 +82,47 @@ def _crate_git_repository_implementation(rctx):
     if result.return_code != 0:
         fail(result.stderr)
 
+    # Recursively check out git submodules to match Cargo's git-dependency
+    # behavior. Cargo always recurses submodules for git sources, so a crate
+    # whose source includes submodules (e.g. khronos_api, which vendors the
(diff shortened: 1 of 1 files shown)


dzbarsky:
Sorry https://github.com/hermeticbuild/rules_rs/pull/113 is gonna conflict with you pretty badly :(

dzbarsky:
Oh actually I think #113 incidentally fixed this, see https://github.com/hermeticbuild/rules_rs/blob/201ceb7a1dc1d86be2d4ed36e6a99ffbd499a6fc/rs/private/git_repository.bzl#L17. Can you double check?

walter-zeromatter:
yep, that fixes it. Thanks!

PR #116, opened 2026-05-02
Add Bazel-native Dylint support

andyscott:
Introduce explicit Dylint rules, nightly rustc-dev plumbing, and integration coverage for target-local custom lint configs.


andyscott:
@dzbarsky 👀

rs/toolchains/declare_rust_analyzer_toolchains.bzl
```diff
@@ -56,17 +57,38 @@ def declare_rust_analyzer_toolchains(
...
-            target_settings = [
-                "@rules_rust//rust/toolchain/channel:" + channel,
-            ],
+            target_settings = target_settings,
             toolchain = rust_analyzer_toolchain_name,
             toolchain_type = "@rules_rust//rust/rust_analyzer:toolchain_type",
             visibility = ["//visibility:public"],
         )
+
+        if toolchain_family_setting != None:
```
dzbarsky, line 79: do we need family on rust-analyzer?

rs/toolchains/declare_rustc_toolchains.bzl
```diff
@@ -160,11 +174,43 @@ def declare_rustc_toolchains(
...
+                        "@rules_rust//rust/toolchain/channel:" + channel,
+                        toolchain_family_setting,
+                    ],
+                    toolchain = rust_toolchain_name,
+                    toolchain_type = "@rules_rust//rust:toolchain_type",
+                    visibility = ["//visibility:public"],
+                )
+
+                native.toolchain(
+                    name = "{}_{}_to_{}_{}_bootstrap_selected_family".format(exec_triple.system, exec_triple.arch, target_key, version_key),
```
dzbarsky, line 202: why do we need extra bootstrap toolchains?

rs/toolchains/declare_rust_analyzer_toolchains.bzl
```diff
@@ -56,17 +57,38 @@ def declare_rust_analyzer_toolchains(
 
         rust_analyzer_toolchain(**rust_analyzer_toolchain_kwargs)
 
+        target_settings = [
+            "@rules_rust//rust/toolchain/channel:" + channel,
```
dzbarsky, line 61: we already set `channel:nightly`, right? Can that take the place of all this `family` plumbing? It's not clear to me what the family is intended to do?

rs/private/rustc_repository.bzl
```diff
@@ -27,8 +27,39 @@ def _symlink_rust_objcopy_shared_libraries(rctx, exec_triple):
 def _rustc_repository_impl(rctx):
     exec_triple = triple(rctx.attr.triple)
     download_and_extract(rctx, "rustc", "rustc", exec_triple)
+    if rctx.attr.rustc_dev_sha256:
```
dzbarsky, line 30: so one of our design goals was to make all downloads fully parallel and lazy. Perhaps rust_dev can be a separate repository and then we link it into the constructed sysroot? (But we can have a "normal" sysroot and a sysroot-with-dev and select between them based on a build flag or something). That would let crates themselves decide which sysroot they want so dylint could opt into the extra one and that would drive the download fully lazily when dylint crate is built? Does that make sense?

dzbarsky:
Apologies for the delay! Looking into it a bit more, I'm not sure if dylint is widely used enough that we should bless it as a first-class extension here, but perhaps putting it in `//examples/dylint` would be a good compromise :) I do think figuring out the rustc-dev story would be good either way and that pretty clearly needs to happen in the toolchain setup, so lets get that bit worked out first!

andyscott:
Can do! Or rather, I'll try 😄.

Perhaps the better play here is to split this up into any needed improvements/adjustments to rules_rs + keep the dylint rule a separate thing outside of rules_rs entirely. I'm certainly open to retaining it in an examples (or experimental?) dir too but I worry it will be a bit "big" for an example.

dzbarsky:
> Can do! Or rather, I'll try 😄.
> 
> Perhaps the better play here is to split this up into any needed improvements/adjustments to rules_rs + keep the dylint rule a separate thing outside of rules_rs entirely. I'm certainly open to retaining it in an examples (or experimental?) dir too but I worry it will be a bit "big" for an example.

Works for me! Probably easier to iterate in a single PR to make sure it works and then split it out before landing?

ags-openai:
Yep, see https://github.com/hermeticbuild/rules_rs/pull/123 with the first bit spun out in https://github.com/hermeticbuild/rules_rs/pull/122.

PR #117, opened 2026-05-03
Patch rules_rust staticlib symbol exports

longlho:
## Summary
- carry the rules_rust `rust_static_library.exported_symbols` patch in rules_rs
- apply that patch from the rules_rust module extension by default
- keep downstream gazelle language modules focused on their local exported-symbol allowlists instead of each stacking the same rules_rust patch

## Verification
- `bazel query @rules_rust//rust/private:all`


longlho:
Closing; this is not the path we want to pursue.

PR #118, opened 2026-05-04
Allow bootstrapping stdlib from source

dzbarsky:


rs/private/build_settings.bzl
```diff
@@ -0,0 +1,7 @@
+def _bool_setting_impl(_ctx):
+    return []
```
cerisier: nothing ?

dzbarsky: fixed :)

PR #132, opened 2026-05-13
Fix binary-only crates

isaacparker0:


rs/rust_crate.bzl
```diff
@@ -139,43 +140,57 @@ def rust_crate(
...
-        target_compatible_with = target_compatible_with,
-        package_metadata = [name + "_package_metadata"],
-        skip_deps_verification = skip_deps_verification,
-        visibility = ["//visibility:public"],
-        skip_per_crate_rustc_flags = True,
-    )
+    if not has_lib:
+        # HACK: create a stub filegroup target so the hub's `<crate>-<version>`
+        # alias (emitted unconditionally in rs/extensions.bzl) still resolves
+        # for binary-only crates. A cleaner fix would be to make the hub skip
```
dzbarsky: yeah i think we'd need to record this in `facts` and that seems not worth it...

dzbarsky (approved):

PR #135, opened 2026-05-18
Flycheck

isaacparker0:


tools/rust_analyzer/BUILD.bazel
```diff
@@ -7,9 +7,11 @@ package(default_visibility = ["//visibility:public"])
 rust_binary(
     name = "discover_bazel_rust_project",
     srcs = ["@rules_rust//tools/rust_analyzer:bin/discover_rust_project.rs"],
+    data = [":flycheck"],
     edition = "2018",
     rustc_env = {
         "ASPECT_REPOSITORY": aspect_repository(),
+        "FLYCHECK_RLOCATIONPATH": "$(rlocationpath :flycheck)",
```
dzbarsky: where is this used?

tools/rust_analyzer/BUILD.bazel
```diff
@@ -19,15 +21,29 @@ rust_binary(
...
+    deps = [
+        "@rrra//:clap",
+        "@rrra//:serde",
+        "@rrra//:serde_json",
     ],
 )
 
 rust_binary(
     name = "gen_rust_project",
     srcs = ["@rules_rust//tools/rust_analyzer:bin/gen_rust_project.rs"],
```
dzbarsky, line 27: i kinda wonder if we should just build it into the same binary and have it re-exec itself instead of building yet another binary?

PR #147, opened 2026-06-15
Sanitize MacOS LLD rpath

### LINKED ISSUE #141: MacOS: prebuilt `rust-lld` ships a leaked CI builder `LC_RPATH` (`/Users/runner/...`)
This is truly a low-prio nit and should be trivial to fix, but I'd like to support rules_rs as much as possible because we are adopting it early. 

## Environment

- macOS 15.6 (24G84), Apple Silicon (arm64)
- Bazel 9.0.2 (via bazelisk), Xcode 17 / Apple clang 17
- `rules_rs` paved-path `MODULE.bazel`, default toolchain `version = "1.92.0"`
  (`rustc 1.92.0 (ded5c06cf 2025-12-08)`)
- Verified on **both** default macOS triples: `aarch64-apple-darwin` and `x86_64-apple-darwin`

### Summary

Thee macOS Rust toolchain that `rules_rs` downloads (stock upstream `static.rust-lang.org` dist)
ships one binary, `rust-lld`, with a hardcoded absolute `LC_RPATH` pointing at the upstream
Rust project's GitHub Actions builder directory:

```
aarch64: /Users/runner/work/rust/rust/build/aarch64-apple-darwin/llvm/lib
x86_64:  /Users/runner/work/rust/rust/build/x86_64-apple-darwin/llvm/lib
```

### Why it Matters

As shipped, this rpath is inert. `rust-lld` has no `@rpath/...` dependencies
(see `otool -L` below), so dyld never consults it. There are two failure paths, and the second we have already experienced:

1. Latent dylib-search hijack. An `LC_RPATH` is a search path for `@rpath/...` dependencies.
   If `rust-lld` ever gains one. A future Rust release makes LLVM a dynamic dep, or the binary is
   patched / re-linked, dyld will search `/Users/runner/work/rust/rust/build/<triple>/llvm/lib`
   **before** `@loader_path/../lib` (it is listed first). On a shared / multi-user macOS host,
   anything able to create that path and plant a dylib there gets it loaded. Narrow preconditions,
   I agree, but only possible because of a stray absolute path.
2. Provenance / audit hygiene. Relocatability checkers, SBOM / SLSA provenance tooling, and
   security scanners flag embedded absolute build paths. Your hermeticity-focused ruleset already
   sanitizes the Linux toolchain (`rs/private/rustc_repository.bzl`) shipping a macOS binary with a
   visible CI builder path is an inconsistency, and it leaks the upstream build environment / layout
   into the artifact.

### Repro

Directly, from the exact tarballs `rules_rs` pins

**Apple Silicon (`aarch64-apple-darwin`):**

```bash
curl -fsSL https://static.rust-lang.org/dist/rustc-1.92.0-aarch64-apple-darwin.tar.xz -o /tmp/rustc-arm.tar.xz
tar -xf /tmp/rustc-arm.tar.xz -C /tmp
otool -l /tmp/rustc-1.92.0-aarch64-apple-darwin/rustc/lib/rustlib/aarch64-apple-darwin/bin/rust-lld | grep -A2 LC_RPATH
```

Output:

```
          cmd LC_RPATH
         path @loader_path/../lib                                              <-- correct, relative
          cmd LC_RPATH
         path /Users/runner/work/rust/rust/build/aarch64-apple-darwin/llvm/lib <--  leaked CI path
```

The issue is present on the other MacOS triple as well. I just got too lazy to add it as a bit under the weather. I can if needed, though. I'd also point out it happens whether direct download or materialized Bazel toolchain. We also did not observe any other absolute rpaths. If you want, I could try to work out a fix unless you know exactly how to solve it.

MarcusSorealheis:
### Problem

Upstream macOS `rust-lld` ships an absolute `LC_RPATH` pointing at the Rust project's CI builder: `/Users/runner/work/rust/rust/build/<triple>/llvm/lib`. It's inert today (`rust-lld` has no `@rpath/...` deps) but leaks the upstream build layout and trips relocatability / SBOM / SLSA provenance scanners.

### Fix

Strip the rpath with a build action keyed to the macOS execution platform, not in the repository rule. A repository rule runs on the machine driving Bazel, which is often a non-macOS scheduler under remote execution, and it cannot re-sign a Mach-O binary. The action runs wherever rust-lld is actually consumed (a dev laptop or a macOS remote-execution worker).

Implemented as a custom rule, consistent with the rest of the ruleset (the codebase uses `ctx.actions` rules and no genrules). Two files:

- `rs/private/sanitize_rust_lld.bzl` (new): a `sanitize_rust_lld` rule. It copies rust-lld and runs `otool` plus `install_name_tool -delete_rpath`. Tools are referenced by absolute path so the action does not depend on the worker's PATH, and arguments are positional so no shell quoting lives in a BUILD file.
- `rs/private/rustc_repository.bzl`: on macOS triples the generated BUILD loads the rule, instantiates it with `exec_compatible_with = ["@platforms//os:macos"]`, and keeps the public rust-lld filegroup interface unchanged (srcs is the sanitized binary, data is the upstream auxiliary linker glob). `include_linker` is suppressed only on macOS.

### Why install_name_tool
Apple Silicon requires every executable to carry a valid code signature, so removing a load command invalidates the binary unless it is re-signed. Apple's `install_name_tool` re-signs the ad-hoc signature automatically, which is why it is used here rather than `@llvm//tools` (which would also force an LLVM download onto every downstream consumer). References:

- install_name_tool man page: https://keith.github.io/xcode-man-pages/install_name_tool.1.html
- Apple Silicon code-signing requirement: https://developer.apple.com/documentation/apple-silicon/porting-your-macos-apps-to-apple-silicon

### DIFF
--- a/rs/private/rustc_repository.bzl
+++ b/rs/private/rustc_repository.bzl
@@ -54,17 +54,65 @@ def _symlink_rust_objcopy_shared_libraries(rctx, exec_triple):
         if entry.basename.startswith("libLLVM"):
             rctx.symlink(entry, "{}/{}".format(rustlib_lib, entry.basename))
 
+# Routes the macOS rust-lld through the sanitize_rust_lld rule (see that file for
+# why this is a build action and not repository-rule work). The public rust-lld
+# filegroup interface is preserved: srcs is the sanitized binary, data carries
+# the auxiliary linker tools unchanged. exec_compatible_with pins the action to a
+# macOS execution platform so it lands on the laptop or macOS RBE worker that
+# actually uses rust-lld.
+_MACOS_RUST_LLD_LOAD = 'load("@rules_rs//rs/private:sanitize_rust_lld.bzl", "sanitize_rust_lld")'
+
+_MACOS_RUST_LLD_BUILD = """\
+sanitize_rust_lld(
+    name = "sanitized-rust-lld",
+    src = "lib/rustlib/{target_triple}/bin/rust-lld{binary_ext}",
+    target_triple = "{target_triple}",
+    exec_compatible_with = ["@platforms//os:macos"],
+)
+
+filegroup(
+    name = "rust-lld",
+    srcs = [":sanitized-rust-lld"],
+    data = glob(
+        include = [
+            "lib/rustlib/{target_triple}/bin/*-ld{binary_ext}",
+            "lib/rustlib/{target_triple}/bin/gcc-ld/*",
+        ],
+        exclude = [
+            "lib/rustlib/{target_triple}/bin/rust-lld{binary_ext}",
+        ],
+        allow_empty = True,
+    ),
+    visibility = ["//visibility:public"],
+)
+"""
+
 def _rustc_repository_impl(rctx):
     exec_triple = triple(rctx.attr.triple)
     download_and_extract(rctx, "rustc", "rustc", exec_triple)
+
     # Upstream Linux rustc bundles libLLVM, which dynamically links against libz.so.1.
     _add_linux_zlib(rctx, exec_triple)
     _symlink_rust_objcopy_shared_libraries(rctx, exec_triple)
-    build_content = [BUILD_for_compiler(
+
+    is_macos = exec_triple.system == "macos"
+
+    # On macOS the linker filegroup is emitted by _MACOS_RUST_LLD_BUILD below so
+    # that rust-lld is routed through the rpath-sanitizing rule; everywhere else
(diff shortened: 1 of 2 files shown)


MarcusSorealheis:
this approach is too complex and we should probably close it.

dzbarsky:
Yeah this is really ugly and I'm not a fan of picking up the run_shell/bash dep. Probably best to fix in rustc upstream if you're motivated :)

PR #149, opened 2026-06-15
Supports non-rustc-built-in `target.cfg.dependencies`

### LINKED ISSUE #148: Support non-rustc-built-in `target.cfg(...).dependencies`
The implementation of rules_rs's target.cfg dependencies resolution not matching cargo's behavior as it only matches the built-in rustc cfgs, the later case is missing

> cfg values come from those built-in to the compiler (run rustc --print=cfg to view) and **extra --cfg flags passed to rustc (such as those defined in RUSTFLAGS)**

https://doc.rust-lang.org/cargo/reference/config.html#target

csmoe:
Fix #148


csmoe:
@dzbarsky may I have a review from you?

dzbarsky:
To be honest I don't understand this PR. It does look like the existing built-in mappings are incomplete, but that seems like it can be handled via fleshing them out rather than needing to compute them on the fly; aren't they immutable?

Then, if you just want to have an extra `--cfg` defined for your build it feels like that should be doable more directly, why do we need to invoke `rustc` for that?

It also seems odd to me to specify a `--cfg` at resolution time; I would expect conceptually that people specify it at build time (though of course in Bazel making it a starlark flag wouldn't help here). To me that suggests that it may not make sense to use this in a Bazel setting, but perhaps you have a use case I'm not imagining that makes it necessary?

csmoe:
> it can be handled via fleshing them out rather than needing to compute them on the fly; aren't they immutable

> why do we need to invoke rustc for that

Because we must confirm whether rustc actually activates the flag, we can optimize this by caching the `--print=cfg` result (as Cargo does), though at least one rustc invocation is always required.
https://github.com/rust-lang/cargo/blob/a595d0da21f228b7fdae64d3d5c0e527ea66bb59/src/cargo/core/compiler/build_context/target_info.rs#L221
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies

> It also seems odd to me to specify a --cfg at resolution time

People might write code like this, using cfg because sdk_dev can't be a feature—doing so would break the additivity of Cargo features. The downside: if --cfg=sdk_dev is missing at resolution time, hack_tls_cert won't be pulled into the dependency graph.
```
target.'cfg(sdk_dev)'.dependenices]
hack_tls_cert = { version = "1.0" }
```

csmoe:
@dzbarsky may I have your review?

dzbarsky:
> People might write code like this, using cfg because sdk_dev can't be a feature—doing so would break the additivity of Cargo features. The downside: if --cfg=sdk_dev is missing at resolution time, hack_tls_cert won't be pulled into the dependency graph.
> 
> ```
> target.'cfg(sdk_dev)'.dependenices]
> hack_tls_cert = { version = "1.0" }
> ```

but `cfg` is also additive. Perhaps this code should instead be using dependency injection or other mechanisms that allow to customize the behavior more dynamically in various ways beyond just features?

More generally, I'm not sold yet that the complexity being introduced here is worth it for all users to unblock the specific use case you have encountered. I'm open to reconsidering if I see more requests for this functionality but right now I think the complexity and overhead of repeated rustc calls (N triples x M closures) may be net negative for most users.

Some of your other issues like https://github.com/hermeticbuild/rules_rs/issues/160 and https://github.com/hermeticbuild/rules_rs/issues/150 make me think that perhaps your use case is a bit more unique than most, and while I understand the reluctance to carry patches, it might be the most pragmatic path for now until we learn more

csmoe:
cfg is not additive as feature's semantics.

If this function is not desired by rules_rs, A warning of non-cargo compatibility should be added to rules_rs's readme or documentation.

PR #151, opened 2026-06-16
Support sharing one crate hub across bzlmod modules via crate.use_hub

### LINKED ISSUE #144: Allow a bzlmod module to consume a `@crates` hub declared by another module
There's no way for one bzlmod module to share a `@crates` hub with another. Every module that uses the `crate` extension must declare its own `from_cargo`, and each `from_cargo` builds an independent hub. So a crate two modules both depend on (e.g. `log`) becomes **two distinct Bazel targets → two rlibs** in the same binary. For any crate with global/`static` state, that breaks the shared-state assumption: `log`'s logger, `tracing`'s dispatcher, cross-crate type identity / `Any`, etc.

Concretely: a repo with a git submodule wired in via `bazel_dep` + `local_path_override` (both real modules in one graph) where the submodule ships a `log` bridge and the parent emits through `log`. Each module gets its own `@crates//:log`, so the logger installed by one is invisible to the other.

(Note: `rules_rust` doesn't support this either [but think that it should](https://github.com/bazelbuild/rules_rust/issues/3157)).

I have an ai assisted workaround which treats a module that uses the crate extension but declares no `from_cargo` as a consumer of a hub another module already built. That way a crate like log resolves to one shared target across the module graph, is this repo open to such a solution? Do you have other ideas of how it might be solved?

pcrumley:
## Problem

A module that uses the `crate` extension must declare its own `crate.from_cargo(...)`, and each `from_cargo` builds an independent `@<name>` hub. So when two modules in the same module graph both depend on a crate (e.g. `log`), it resolves to **two distinct Bazel targets** — two rlibs linked into the same binary. For crates with process-global `static` state this breaks the shared-state assumption: `log`'s logger, `tracing`'s global dispatcher, and anything relying on cross-crate type identity / `Any` downcasting.

Resolves [https://github.com/hermeticbuild/rules_rs/issues/144](https://github.com/hermeticbuild/rules_rs/issues/144)

## What this changes

Add a `crate.use_hub(name = ...)` tag, where `name` is a hub another module declared (e.g. `crates`). A module that points at an existing hub instead of declaring its own `from_cargo` shares that hub, so any crate in it (e.g. `log`) resolves to a single target across both modules instead of one copy per module.

```python
# Owning module (e.g. the root) — declares the hub:
crate.from_cargo(name = "crates", cargo_toml = "//:Cargo.toml",
                 cargo_lock = "//:Cargo.lock", platform_triples = [...])
use_repo(crate, "crates")
```

```python
# Consumer module (e.g. a git submodule with its own MODULE.bazel):
crate.use_hub(name = "crates")   # share the owning module's hub
use_repo(crate, "crates")
```

`@crates//:log` is now one target shared across both modules.

### Behavior details

- A module that uses the `crate` extension must declare either `from_cargo` or `use_hub`. A `use_hub` that references a hub no module creates fails with a clear message listing the available hubs. Existing users see no behavior change.
- A module that is built both standalone and as a dependency can pair `use_hub` with its own `from_cargo` declared as a `dev_dependency`. It builds its own hub when built standalone and shares the owning module's hub when built as a dependency. In the standalone case the hub is created by the dev `from_cargo` but imported via the regular `use_hub`, so the regular `use_hub` reports it as a *regular* root direct dep (rather than a dev dep) to avoid a bzlmod import-mismatch warning.
- A `use_hub` module triggers no repo generation by the `crate` extension. The hub (and its spoke repos) are created once, by the owning module's `from_cargo`, and `use_hub` only imports that existing hub. 
- The pre-existing "two `from_cargo` with the same name collide" error is unchanged.

### Known limitations

Because rules_rs has only one hub, the dependency has to reference something the parent provides, the hub name. This can be a pretty generic name, here it suggests 'crates' so its a relatively loose coupling. However its a bit strange, and I would understand if the maintainers are not amenable to this solution. However its a relatively small change that is low risk, and it also closes an issue with what I've seen has no solution in the bazel rust ecosystem, including `rules_rust`.

## Implementation

All in `rs/extensions.bzl`:
- New `_use_hub` tag class (single `name` attr, default `"crates"`), registered on the `crate` module extension.
- The per-module `from_cargo`-required check becomes "`from_cargo` **or** `use_hub`", preserving a precise per-module error.
- Validation that every `use_hub` references a hub some module creates.
- Root-module direct-dep bookkeeping so a consumer's `use_repo` of a foreign hub is reported correctly (regular `use_hub` wins over a same-named dev `from_cargo`).

## Test

A red-green  test exists in `test/shared_hub/` (hub owner with a `log`-installing logger crate) + `test/shared_hub_child/` (a separate bzlmod module that consumes the hub via`crate.use_hub`). The `rust_test //:shared_hub_sharing_test` asserts at runtime that a `log` record emitted by the consumer module reaches the logger installed by the owner.

## Docs

README gains a "Sharing one hub across bzlmod modules" section covering the `use_hub` consumer pattern, the `dev_dependency` standalone idiom, and the `all_crate_deps()` limitation.


csmoe:
Just for a note, what about the features resolution? I mean module A use `log = { version = "1", features = "foo"}`, and module B expect log with `feature = "bar"`, since `use_hub` will not regen the repo, then what will happend in this case.

pcrumley:
this is a great point. the `use_hub` as implemented doesn't have the ability to set any features, because the only reasonable thing to do is just compile in whatever the hub would have (it's outside of the use_hub control what hub it actually grabs other than its name). This did make me think and it seems reasonable to have the `use_hub` api have an `expect_features` option, to make sure that the hub has those features & throw a reasonable error. (update to this PR forthcoming)

*edit to add* I also added expect_rev & expect_version

pcrumley:
I'm going to close this, because while it does work, there are some things I am unhappy with in the API, like the child needing to know about the parent. In addition, its kind of unclear how much normal users would hit this when following best practices in how to have a multicrate bazel monorepo.

I'll leave it up to the maintainers if they want to also close the corresponding issue: https://github.com/hermeticbuild/rules_rs/issues/144

Issue #153, opened 2026-06-16
ARM `eabi`/`eabihf` triples are unbuildable: "Illegal ambiguous match" on `target_triple`


pcrumley:
The generated toolchains `select()` on per-triple `config_setting`s whose constraints (`cpu`, `os`, libc) come from `triple_to_rust_constraint_set()` in `rs/platforms/triples.bzl`. That projection carries no ARM float-ABI bit, so soft-float and hard-float triples collide on an identical constraint set and Bazel fails:

```
Illegal ambiguous match on configurable attribute "target_triple"
@@rules_rs+//rs/platforms/config:arm-unknown-linux-musleabi
@@rules_rs+//rs/platforms/config:arm-unknown-linux-musleabihf
```

There are 7 colliding groups today:

| Constraint set | Triples |
|---|---|
| `cpu:arm, os:linux, glibc` | `arm-unknown-linux-gnueabi` / `gnueabihf` |
| `cpu:armv7, os:linux, glibc` | `armv7-unknown-linux-gnueabi` / `gnueabihf` |
| `cpu:arm, os:linux, musl` | `arm-unknown-linux-musleabi` / `musleabihf` |
| `cpu:armv7, os:linux, musl` | `armv7-unknown-linux-musleabi` / `musleabihf` |
| `cpu:armv8-m, os:none` | `thumbv8m.main-none-eabi` / `eabihf` |
| `cpu:aarch64, os:none` | `aarch64-unknown-none` / `aarch64-unknown-none-softfloat` |
| `cpu:wasm32, os:wasi, wasi_preview_1` | `wasm32-wasip1` / `wasm32-wasip1-threads` |

(`thumbv7em-none-eabi`/`eabihf` do *not* collide: `rules_rust` already maps them to distinct CPUs, `armv7e-m` vs `armv7e-mf`.)

Every supported triple should project to a unique constraint set so toolchain resolution is unambiguous, without changing existing zero-config builds.

Repro: `bazel build //some:target --platforms=@rules_rs//rs/platforms:arm-unknown-linux-musleabihf`

dzbarsky:
Thank for fixing!

PR #154, opened 2026-06-16
Disambiguate ARM hard-float vs soft-float (and WASM threads) target triples

pcrumley:
## Problem

The generated toolchains `select()` on per-triple `config_setting`s whose constraints came from `triple_to_rust_constraint_set()` — `cpu`, `os`, libc, but no float-ABI or WASM-threads bit. So 7 groups of triples projected to identical constraint sets (the `arm`/`armv7` × `gnu`/`musl` `eabi`/`eabihf` pairs, `thumbv8m.main-none-eabi`/`eabihf`, `aarch64-unknown-none`/`-softfloat`, and `wasm32-wasip1`/`wasm32-wasip1-threads`), and building for a matching platform failed with `Illegal ambiguous match on configurable attribute "target_triple"`.

## Fix

New constraints in `rs/platforms/constraints/BUILD.bazel`: a `float_abi` setting (`hardfloat` default / `softfloat`) and a `wasm_threads` setting (`wasm_threads_off` default / `wasm_threads_on`). The defaults are the conventional/plain variant so existing platforms resolve unchanged; the other variant opts in explicitly.

`triple_to_rust_constraint_set` (`rs/platforms/triples.bzl`) appends a disambiguator only when a triple has a sibling that is also in the target set *and* projects to the same base constraint set — a genuine collision. This covers the `eabi`↔`eabihf`, `-softfloat`-suffix, and `-threads`-suffix encodings. Scoping to actual base-set equality leaves lone targets untouched and skips pairs that already differ (e.g. `thumbv7em-none-eabi`/`eabihf`, which `rules_rust` already maps to distinct CPUs), so non-annotated consumers keep resolving. The `targets` list is threaded through to the config-settings, platforms, and toolchain projections so a custom triple subset stays self-consistent. No change to the `select()` construction is needed — unique config settings disambiguate on their own.

New Starlark suite in `rs/platforms/triples_test.bzl` asserts the uniqueness invariant over all triples, that each colliding pair is separated by a distinct disambiguator, and that lone targets gain none. README "Cross ABI target details" documents the new constraints.

## Verification

`//rs/platforms:all` and `//rs/private:all` tests pass. `@default_rust_toolchains//:all` loads with distinct `aarch64_unknown_none` vs `_softfloat` toolchains. `bazel cquery` across all eight previously-colliding platforms resolves cleanly with no ambiguous match. `config:arm-unknown-linux-musleabi` → `[…, musl, softfloat]` vs `musleabihf` → `[…, musl, hardfloat]`. Non-ARM/non-WASM config settings and platforms are unaffected. No triples removed and no consumer-facing tags/attributes changed.

proposed fix for [https://github.com/hermeticbuild/rules_rs/issues/153](https://github.com/hermeticbuild/rules_rs/issues/153)


rs/platforms/triples.bzl
```diff
@@ -24,8 +30,72 @@ def triple_to_rust_constraint_set(target_triple):
...
+    constraints = _base_rust_constraint_set(target_triple)
+
+    disambiguator = _disambiguator_constraint(target_triple, targets or ALL_TARGET_TRIPLES)
+    if disambiguator:
+        constraints.append(disambiguator)
+
+    return constraints
+
+def triple_to_constraint_set(target_triple, targets = None):
+    constraints = triple_to_rust_constraint_set(target_triple, targets)
```
dzbarsky, line 80: I don't understand why the mechanism above is different from the one below, can we just follow the pattern below, it seems easier to reason about?

rs/platforms/triples_test.bzl
```diff
@@ -0,0 +1,118 @@
...
+    "@rules_rs//rs/platforms/constraints:wasm_threads_on",
+    "@rules_rs//rs/platforms/constraints:wasm_threads_off",
+]
+
+# Triples that have no colliding sibling and so must NOT gain a disambiguator:
+# bare/lone thumb and android eabi targets, a lone netbsd eabihf target, and a
+# representative non-ARM triple. thumbv7em-none-{eabi,eabihf} are included
+# because rules_rust already maps them to distinct CPUs (armv7e-m vs armv7e-mf),
+# so they do not collide and tagging them would break non-annotated consumers.
+_LONE_TRIPLES = [
```
dzbarsky: I wonder if we could just have a simple rust target that we transition to all target triples as more of an integration test vs this testing internals?

dzbarsky:
@pcrumley Hello again :) My apologies for this papercut - it's a gap I've been aware of but was putting off because the right home for these constraints is in https://github.com/bazel-contrib/platforms_contrib which is getting actively built out. (cc @fmeum) But since you've hit this, I think it's fine if we fix it here and later alias to contrib + deprecate these constraints; it's not the end of the world

Could you take a look at my suggestions and check if it's doable to simplify?

rs/platforms/triples_test.bzl
pcrumley: I thought it would be too heavy to have to pull in a toolchain for every target platform, and for some of the platforms you get cc toolchain errors for even very simple targets. I think i figured out a way that still actually iterates over all of triples and doesn't just sort of test the own internal logic of it. This test is still a red-green test from this change

rs/platforms/triples.bzl
pcrumley: yeah the reason is you do need to call out to a helper to see if there is a sibling triple, however, you can keep the inline in here and only have at sibling logic be a helper.

PR #156, opened 2026-06-17
Add opt_level to toolchain

marcgtg:
Adds opt_level to toolchain declaration to allow overruling the defaults defined in rust_rules

Pre-commit hooks required some additional changes.

### DIFF
--- a/MODULE.bazel.lock
+++ b/MODULE.bazel.lock
@@ -398,6 +398,69 @@
         }
       }
     },
+    "@@llvm+//3rd_party/gcc/extension:gcc.bzl%gcc": {
+      "general": {
+        "bzlTransitiveDigest": "XUWTels9SzO7xhmQBqYtaFYFMTjrFMSDFi/Q7hv4f34=",
+        "usagesDigest": "9IP/e++b81Ga1zICxzXu+yqfe5i4DX5F/SsrgV5a568=",
+        "recordedInputs": [
+          "REPO_MAPPING:llvm+,bazel_lib bazel_lib+",
+          "REPO_MAPPING:llvm+,bazel_tools bazel_tools"
+        ],
+        "generatedRepoSpecs": {
+          "gcc": {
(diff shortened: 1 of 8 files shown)


dzbarsky:
Thanks for the PR! I'm wondering why overriding the opt-level via CLI flag (or transition) doesn't work, can you help me understand your use case?

marcgtg:
I'm not sure I understand the question correctly. I can also pass the opt-level via the extra_rustc_flag like this:

`build --@rules_rust//rust/settings:extra_rustc_flag=-Copt-level=1`

But after doing some investigation we were afraid that this might conflict with the opt_level setting in rust_rules. Hence the changes.

We need the opt_level because we have heavy arithmetic operations, and with opt-level=0 the executable is about 50x slower which would affect debugging use cases.

If there is a simpler solution I would like to hear about it.

dzbarsky:
Yes, exactly, I was wondering if you can just use `--@rules_rust//rust/settings:extra_rustc_flag=-Copt-level=1`. Are you saying that the toolchain opt_level overrode it? (If so, that's a bug IMO and we should fix it).

Your desire to optimize numeric code makes sense, but I feel like wanting to override the optimization level should be a decision at the crate level (or global one), doing it on the toolchain seems a bit odd. Would you really want it to vary based on the platform the compiler executes on?

marcgtg:
In .bazelrc I tried:

build --@rules_rust//rust/settings:extra_rustc_flag=-Copt-level=1
build:release --@rules_rust//rust/settings:extra_rustc_flag=-Copt-level=3

And that did work, so it wasn't overruled, but we were concerned that this would not be the correct approach seeing that there is a specific opt-level setting in rules_rust

But you're right that it doesn't make much sense to have different settings for different toolchains. If you're saying that extra_rustc_flags is the correct way to go than this pull request can be rejected.

dzbarsky:
Yes, this is the correct approach in my opinion. If it's working for you, let's close this out!

PR #158, opened 2026-06-22
[codex] Split build-script feature resolution

dzbarsky:
Cargo resolver v2 keeps normal-dependency features separate from build-dependency features, but `rules_rs` previously resolved both dependency kinds into one external `rust_crate` target. This change resolves target dependencies without build dependencies, seeds a second resolution over `SUPPORTED_EXEC_TRIPLES`, and generates either one target when both resolutions match or `_target` and `_exec` targets selected by `@rules_rust//cargo/settings:use_exec_features_enabled` when they differ. It pins `rules_rust` commit [`7837073014b566291368027aa116cd03fe108049`](https://github.com/hermeticbuild/rules_rust/commit/7837073014b566291368027aa116cd03fe108049), which composes the build-script exec transition with the marker transition through `config.exec().and_then(...)`.

The `build_exec_feature_split` fixture requires `target_feature` from a normal dependency and `exec_feature` from the same crate as a build dependency. The shared crate fails compilation if both features are enabled, so the fixture verifies that the target library and build script receive different feature sets.

Validation used source-built Bazel commit `0878dd724149b73a731ab83859a55a65e9b438ed`: `@build_exec_feature_split//:_workspace_deps` passed with 1,720 actions, `//rs/private:all` passed 28/28 tests, `cargo check --locked` passed for the fixture, and Buildifier plus `git diff --check` passed.

Current scope: proc macros are unchanged, and `seed_exec_build_dependencies` unions features from all active target triples before resolving each exec triple. Exact target-specific build-script features require resolution keyed by both target triple and exec triple.


pawelchcki:
I’m hitting what looks like relvant issue while building ESP32-C6 firmware with `rules_rs` 0.0.102.                                                                                        


  ```starlark
  platform_triples = [
      "riscv32imac-unknown-none-elf",
      "x86_64-unknown-linux-gnu",
  ]
  ```

  `portable-atomic` is compiled as a host tool with a target-only feature:

  ```text
  --target=x86_64-unknown-linux-gnu
  --cfg feature="unsafe-assume-single-core"
  ```

  This fails because `unsafe-assume-single-core` is unsupported on x86_64.

FYI  I also see leakage in the other direction: `log` is compiled for bare-metal RISC-V with `feature="std"`, despite the applicable dependency disabling default features.

dzbarsky:
@pawelchcki does this PR fix the issue for you? It feels like the real problem may be crates enabling features in their deps unconditionally instead of conditioning on cfg-target - but hopefully this helps. We can land it if it does just need to rebase and add a regression test, etc. let me know!

zbarsky-openai:
This introduces a crate-identity mismatch when a build script depends on a first-party workspace library. A minimal three-crate reproducer passes with Cargo and the PR base, but fails at `931bb29200bcebaea1f4a9f817d4df0b2a52e566`, including when the caller uses `cargo_build_script` from `@cargo//:defs.bzl`.

The root crate normally depends on `shared` with `target_feature`. Its build script depends on `shared` with `exec_feature` and on a workspace library, `helper`. `helper::token()` returns `shared::Token`, and the build script passes that value to `fn accept(_: shared::Token)`. `Token` is an empty struct; neither feature changes its API.

The generated dependency labels explain the failure:

```text
build script -> //helper -> @cargo//:shared-0.1.0
build script -----------> @cargo//:__exec/shared-0.1.0_exec
```

```text
error[E0308]: mismatched types
 --> build.rs:3:20
  |
3 | fn main() { accept(helper::token()); }
  |             ------ ^^^^^^^^^^^^^^^ expected `Token`, found `shared::Token`

note: there are multiple different versions of crate `shared` in the dependency graph
```

| Control | Result |
| --- | --- |
| Cargo 1.95.0, resolver 2 | Pass |
| Bazel with PR base `4853cceb2fae09e101ae75fb2251b022fb8662fc` | Pass |
| Same Bazel with PR head `931bb29200bcebaea1f4a9f817d4df0b2a52e566` | E0308 above |
| Same PR head, changing only `helper` from a workspace member to an excluded path dependency | Pass |

In the last control, rules_rs generates `helper` itself and gives it the same execution variant of `shared` as the build script. The missing case is propagation through first-party workspace libraries, whose BUILD rules still use `all_crate_deps(normal = True)`.

<details>
<summary>Self-contained reproducer: save as create-repro.py in an empty directory</summary>

```python
from pathlib import Path
import sys

variant = sys.argv[1]
revision, integrity = {
    "baseline": ("4853cceb2fae09e101ae75fb2251b022fb8662fc", "sha256-V4a6hp4ftqVpItMAyP1zfI9oxraMlrwBDbXJiUcMCGU="),
    "candidate": ("931bb29200bcebaea1f4a9f817d4df0b2a52e566", "sha256-0RDrd8jfI885oyLo0z6hF+t5oYZjYj/v0MipcTDsLvA="),
}[variant]
files = {
    'Cargo.toml': '''[package]
name = "workspace-build-identity"
version = "0.1.0"
edition = "2021"

[workspace]
resolver = "2"
members = ["helper"]
exclude = ["vendor/shared"]

[dependencies]
shared = { path = "vendor/shared", features = ["target_feature"] }

[build-dependencies]
helper = { path = "helper" }
shared = { path = "vendor/shared", features = ["exec_feature"] }
''',
    'build.rs': '''fn accept(_: shared::Token) {}

fn main() { accept(helper::token()); }
''',
    'src/lib.rs': '''pub fn token() -> shared::Token { shared::Token }
''',
    'helper/Cargo.toml': '''[package]
name = "helper"
version = "0.1.0"
edition = "2021"

[dependencies]
shared = { path = "../vendor/shared" }
''',
    'helper/BUILD.bazel': '''load("@cargo//:defs.bzl", "all_crate_deps")
load("@rules_rs//rs:rust_library.bzl", "rust_library")

rust_library(
    name = "helper",
    srcs = ["src/lib.rs"],
    edition = "2021",
    deps = all_crate_deps(normal = True),
)
''',
    'helper/src/lib.rs': '''pub fn token() -> shared::Token { shared::Token }
''',
    'vendor/shared/Cargo.toml': '''[package]
name = "shared"
version = "0.1.0"
edition = "2021"

[features]
exec_feature = []
target_feature = []
''',
    'vendor/shared/src/lib.rs': '''pub struct Token;
''',
    'skip-unused-tool-cargo.patch': '''--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -84,1 +84,1 @@
-crate = use_extension("//rs:extensions.bzl", "crate")
+crate = use_extension("//rs:extensions.bzl", "crate", dev_dependency = True)
''',
    'MODULE.repro.bazel': '''module(name = "workspace_build_identity")
bazel_dep(name = "rules_rs", version = "0.0.0")
archive_override(
    module_name = "rules_rs",
    integrity = "INTEGRITY",
    patches = ["//:skip-unused-tool-cargo.patch"],
    patch_strip = 1,
    strip_prefix = "rules_rs-REVISION",
    urls = ["https://github.com/hermeticbuild/rules_rs/archive/REVISION.tar.gz"],
)
bazel_dep(name = "llvm", version = "0.8.18")
register_toolchains("@llvm//toolchain:all")

crate.from_cargo(
    name = "cargo",
    cargo_lock = "//:Cargo.lock",
    cargo_toml = "//:Cargo.toml",
    platform_triples = ["x86_64-unknown-linux-gnu"],
)
use_repo(crate, "cargo")
toolchains = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")
toolchains.toolchain(edition = "2021", version = "1.95.0")
use_repo(toolchains, "default_rust_toolchains")
register_toolchains("@default_rust_toolchains//...")
''',
    'BUILD.baseline.bazel': '''load("@cargo//:defs.bzl", "aliases", "all_crate_deps")
load("@rules_rs//rs:cargo_build_script.bzl", "cargo_build_script")

cargo_build_script(
    name = "build_script",
    srcs = ["build.rs"],
    aliases = aliases(),
    edition = "2021",
    deps = all_crate_deps(build = True),
)
''',
    'BUILD.candidate.bazel': '''load("@cargo//:defs.bzl", "cargo_build_script")

cargo_build_script(
    name = "build_script",
    srcs = ["build.rs"],
)
''',
}
for name, content in files.items():
    path = Path(name)
    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text(content)
Path("MODULE.bazel").write_text(
    files["MODULE.repro.bazel"].replace("REVISION", revision).replace("INTEGRITY", integrity)
)
Path("BUILD.bazel").write_text(files[f"BUILD.{variant}.bazel"])

```

</details>

Run on Linux x86_64 with Rust/Cargo 1.95.0. The verified Bazel runs used `9.3.0-dzbarsky34`, sandboxed execution, and fresh servers between revisions. The public helper does not exist at the PR base, so the baseline BUILD uses the existing `cargo_build_script` plus `all_crate_deps(build = True)` and `aliases()`.

```bash
python3 create-repro.py baseline
cargo generate-lockfile --offline
cargo build --offline --locked
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:build_script \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown

python3 create-repro.py candidate
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:build_script \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
# Expected: E0308.
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown

# Change only workspace membership; keep all Rust and BUILD source unchanged.
python3 - <<'PY'
from pathlib import Path
p = Path("Cargo.toml")
p.write_text(p.read_text().replace(
    'members = ["helper"]\nexclude = ["vendor/shared"]',
    'members = []\nexclude = ["helper", "vendor/shared"]',
))
PY
cargo generate-lockfile --offline
cargo build --offline --locked
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:build_script \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
# Expected: pass.
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown
```

Setup caveat: both Bazel revisions received the same diagnostic MODULE patch, included above, marking rules_rs's own Cargo extension `dev_dependency = True`. This avoided unrelated tool-repository downloads that failed TLS on the test host. The reproducer's Cargo extension remains enabled, and the resolver/rule implementations are unchanged. These are correctness controls, not performance measurements.

The fix needs to select execution dependencies consistently through first-party library edges while retaining those libraries' source inputs and rule customizations. Using the public build-script helper alone does not cover that boundary.

zbarsky-openai:
There is a second regression when Bazel explicitly builds a generated crate that is inactive in Cargo's selected-platform graph: the target is still emitted, but its mandatory build dependency is dropped.

A three-crate example declares `optional-crate` as a Windows-only dependency of the root. The dependency itself is platform-independent, and its build script unconditionally calls `build_helper::configure()`. The Cargo extension resolves Linux only; Bazel then explicitly builds `@cargo//:optional-crate-0.1.0` on Linux.

| Control | Result |
| --- | --- |
| Cargo 1.95.0 builds the dependency directly through its manifest | Pass |
| Bazel at PR base `4853cceb2fae09e101ae75fb2251b022fb8662fc` | Pass |
| Same Bazel at PR head `931bb29200bcebaea1f4a9f817d4df0b2a52e566` | E0433: missing `build_helper` |
| Same PR head, changing only the root dependency condition from Windows to Linux | Pass |

The base-generated rule has:

```starlark
build_deps = ["@cargo//:build-helper-0.1.0"]
```

The PR-generated rule retains `build_script = "build.rs"` and allows the Linux platform, but its selected script definition has `deps = []` and `deps_by_platform = {}`. Rust reports:

```text
error[E0433]: cannot find module or crate `build_helper` in this scope
 --> .../build.rs:1:13
  |
1 | fn main() { build_helper::configure(); }
  |             ^^^^^^^^^^^^ use of unresolved module or unlinked crate `build_helper`
```

Changing only the root condition to Linux restores `@cargo//:build-helper-0.1.0` in the script definition and makes the same target compile.

This follows the inactive-package path: `collect_exec_build_dependencies` skips inactive owners, while `prepare_dependency_variants` still emits a fallback definition when there are no active target or execution nodes. That fallback lacks the build dependency. Explicitly requesting a generated crate from Bazel needs to retain its dependency closure even when the Cargo root does not activate it for that platform.

<details>
<summary>Self-contained reproducer: save as create-repro.py in a separate empty directory</summary>

```python
from pathlib import Path
import sys

variant = sys.argv[1]
revision, integrity = {
    "baseline": ("4853cceb2fae09e101ae75fb2251b022fb8662fc", "sha256-V4a6hp4ftqVpItMAyP1zfI9oxraMlrwBDbXJiUcMCGU="),
    "candidate": ("931bb29200bcebaea1f4a9f817d4df0b2a52e566", "sha256-0RDrd8jfI885oyLo0z6hF+t5oYZjYj/v0MipcTDsLvA="),
}[variant]
files = {
    'Cargo.toml': '''[package]
name = "inactive-build-repro"
version = "0.1.0"
edition = "2021"

[workspace]
resolver = "2"
exclude = ["vendor/optional-crate", "vendor/build-helper"]

[target.'cfg(target_os = "windows")'.dependencies]
optional-crate = { path = "vendor/optional-crate" }

[patch.crates-io]
build-helper = { path = "vendor/build-helper" }
''',
    'skip-unused-tool-cargo.patch': '''--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -84,1 +84,1 @@
-crate = use_extension("//rs:extensions.bzl", "crate")
+crate = use_extension("//rs:extensions.bzl", "crate", dev_dependency = True)
''',
    'MODULE.repro.bazel': '''module(name = "inactive_build_repro")
bazel_dep(name = "rules_rs", version = "0.0.0")
archive_override(
    module_name = "rules_rs",
    integrity = "INTEGRITY",
    patches = ["//:skip-unused-tool-cargo.patch"],
    patch_strip = 1,
    strip_prefix = "rules_rs-REVISION",
    urls = ["https://github.com/hermeticbuild/rules_rs/archive/REVISION.tar.gz"],
)
bazel_dep(name = "llvm", version = "0.8.18")
register_toolchains("@llvm//toolchain:all")

crate.from_cargo(
    name = "cargo",
    cargo_lock = "//:Cargo.lock",
    cargo_toml = "//:Cargo.toml",
    platform_triples = ["x86_64-unknown-linux-gnu"],
)
use_repo(crate, "cargo")
toolchains = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")
toolchains.toolchain(edition = "2021", version = "1.95.0")
use_repo(toolchains, "default_rust_toolchains")
register_toolchains("@default_rust_toolchains//...")
''',
    'BUILD.bazel': '''alias(name = "optional_crate", actual = "@cargo//:optional-crate-0.1.0")
''',
    'src/lib.rs': '''pub fn root() {}
''',
    'vendor/optional-crate/Cargo.toml': '''[package]
name = "optional-crate"
version = "0.1.0"
edition = "2021"

[build-dependencies]
build-helper = { path = "../build-helper" }
''',
    'vendor/optional-crate/build.rs': '''fn main() { build_helper::configure(); }
''',
    'vendor/build-helper/Cargo.toml': '''[package]
name = "build-helper"
version = "0.1.0"
edition = "2021"
''',
    'vendor/optional-crate/src/lib.rs': '''pub fn optional() {}
''',
    'vendor/build-helper/src/lib.rs': '''pub fn configure() {}
''',
}
for name, content in files.items():
    path = Path(name)
    path.parent.mkdir(parents=True, exist_ok=True)
    path.write_text(content)
Path("MODULE.bazel").write_text(
    files["MODULE.repro.bazel"].replace("REVISION", revision).replace("INTEGRITY", integrity)
)

```

</details>

Run on Linux x86_64 with Rust/Cargo 1.95.0. The verified runs used Bazel `9.3.0-dzbarsky34`, sandboxed execution, and a fresh server for each control.

```bash
python3 create-repro.py baseline
cargo generate-lockfile --offline
cargo generate-lockfile --offline --manifest-path=vendor/optional-crate/Cargo.toml
cargo build --offline --locked --manifest-path=vendor/optional-crate/Cargo.toml
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:optional_crate \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown

python3 create-repro.py candidate
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:optional_crate \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
# Expected: E0433.
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown

python3 - <<'PY'
from pathlib import Path
p = Path("Cargo.toml")
p.write_text(p.read_text().replace('target_os = "windows"', 'target_os = "linux"'))
PY
bazel --nosystem_rc --nohome_rc --noworkspace_rc build //:optional_crate \
  --lockfile_mode=update --spawn_strategy=sandboxed --jobs=2
# Expected: pass.
bazel --nosystem_rc --nohome_rc --noworkspace_rc shutdown
```

As in the first reproducer, both revisions use the same diagnostic MODULE patch marking rules_rs's own Cargo extension `dev_dependency = True`, to avoid unrelated tool-repository downloads that fail TLS on the test host. The fixture's Cargo extension and the resolver/rule implementations are unchanged. These are correctness controls, not performance measurements.

rs/patches/rules_rust_cargo_context.patch
```diff
@@ -0,0 +1,343 @@
...
++    inputs = [],
++    outputs = [CARGO_CONTEXT_SETTING],
++)
++
+ rust_toolchain = rule(
+     implementation = _rust_toolchain_impl,
++    cfg = _clear_cargo_context,
+     fragments = ["cpp"],
+     attrs = {
++        "_allowlist_function_transition": attr.label(
```
dzbarsky: snip this

rs/patches/rules_rust_cargo_context.patch
```diff
@@ -0,0 +1,343 @@
...
+ load("//rust/private:lto.bzl", "RustLtoInfo")
++load("//rust/private:per_crate_flag_trim.bzl", "CARGO_CONTEXT_SETTING")
+ load(
+     "//rust/private:rust_allocator_libraries.bzl",
+     "make_libstd_and_allocator_ccinfo",
+@@ -699,10 +700,23 @@
+         make_variable_info,
+     ]
+ 
++def _clear_cargo_context_impl(_settings, _attr):
```
dzbarsky: is "cargo_context" really the best name? seems kinda slop

README.md
```diff
@@ -164,7 +164,7 @@ load("@rules_rs//rs:rust_library.bzl", "rust_library")
 rust_library(
     name = "lib",
     srcs = ["src/lib.rs"],
-    aliases = aliases(),
+    aliases = aliases(normal = True),
```
dzbarsky: this shouldn't be necessary, did we lose our nice defaulting...

README.md
```diff
@@ -484,6 +484,65 @@ use_repo(rules_rust_pyo3, "rules_rust_pyo3")
...
+`platform_triples` entry. Each generated crate has one library target, with
+features and dependencies selected by Cargo resolution. Build resolutions can
+share the default Bazel configuration when features, transitive dependencies,
+and build-script requirements agree. Otherwise, they retain the original target
+triple; equivalent build resolutions do not substitute another target triple.
+Crates unreachable from the Cargo roots on every configured platform are
+incompatible. Generating a Bazel label does not make the crate an additional
+Cargo root.
+
+Build scripts are grouped by target features, build dependencies, and aliases.
```
dzbarsky: do we actually need to group by aliases? I think those are just a consequence of build deps no?

README.md
```diff
@@ -484,6 +484,65 @@ use_repo(rules_rust_pyo3, "rules_rust_pyo3")
...
+load("@crates//:defs.bzl", "cargo_build_script")
+
+cargo_build_script(
+    name = "build_script",
+    srcs = ["build.rs"],
+    crate_root = "build.rs",
+)
+```
+
+Use `all_crate_deps(normal = True)`, `aliases(normal = True)`, and
```
dzbarsky: all of these shouldn't need normal = True by default!

README.md
```diff
@@ -484,6 +484,65 @@ use_repo(rules_rust_pyo3, "rules_rust_pyo3")
...
+shared. Cargo registry metadata does not include these Bazel dependencies.
+
+Proc macros reached through normal dependencies use the existing conservative
+target feature resolution for their normal dependencies. They do not share
+features enabled only through build dependencies, so they may need explicit
+`crate_features` annotations.
+For example, a PyO3 toolchain that sets `PYO3_NO_PYTHON` needs its chosen
+`abi3-py3*` feature on `pyo3-build-config` in both resolutions.
+
+`gen_binaries` makes an otherwise build-only package a target root with its
```
dzbarsky: didn't we remove target roots?

zbarsky-openai:
There is a remaining compatibility case for a standalone `rust_binary` whose dependencies come from the package's `[build-dependencies]`. This is an executable invoked by another build rule, rather than a Cargo build-script runner; switching its rule to `cargo_build_script` changes that interface.

Using the three-crate fixture from [the earlier reproducer](https://github.com/hermeticbuild/rules_rs/pull/158#issuecomment-5770236512), replace the root `BUILD.bazel` with:

```starlark
load("@cargo//:defs.bzl", "all_crate_deps")
load("@rules_rs//rs:rust_binary.bzl", "rust_binary")

rust_binary(
    name = "build_tool",
    srcs = ["build.rs"],
    edition = "2021",
    deps = all_crate_deps(build = True),
)
```

Run `bazel build //:build_tool`:

| rules_rs revision | Result |
| --- | --- |
| PR base `4853cceb2fae09e101ae75fb2251b022fb8662fc` | Builds successfully |
| PR head `b49cd17e253228129bbe58f196f63714427dd920` | Fails during package loading |

```text
Build dependencies require a different Cargo resolution. Use cargo_build_script from the generated Cargo repository's defs.bzl.
```

For the second run, change the fixture's archive revision/strip prefix to `b49cd17e253228129bbe58f196f63714427dd920` and its integrity to `sha256-K0MznCeolqE31to9cvEgnzh4Rh4G8uQiBDU0eE8aXD4=`. The Cargo manifests, source, and lockfile are identical in both controls. Both use Bazel 9.3.0-dzbarsky34; the same diagnostic patch marking the unused developer-tool Cargo extension as a dev dependency remains applied.

The build-script helper now carries the Cargo resolution through first-party dependencies. A custom executable needs access to that same resolution and transition without going through the build-script runner.

Separately, I traced the inactive-crate consumer behind my second reproducer: it lacked a declaration in the Cargo graph being imported. Fixing that consumer is appropriate; an explicitly incompatible target is a clearer failure than the earlier missing build dependency.

zbarsky-openai:
`workspace_dep_data()` drops the `binaries` and `shared_libraries` fields on the current head, `dce9b7ee46cd87cb2dcf679993f5341685e9f74a`. Both are emitted at the PR base, `4853cce`, by extracting `bin` and `cdylib` entries from Cargo's `targets` metadata and making their source paths package-relative.

Consumers that generate first-party targets from this metadata now lose their binaries, or fail when validating explicit binary source declarations. These target declarations are independent of the new dependency-resolution configurations, so retaining the existing extraction and fields should preserve that contract.

Minimal reproduction: add these inputs and assertions to the existing `workspace_dep_data_tests` fixture:

```diff
diff --git a/rs/private/workspace_dep_data_test.bzl b/rs/private/workspace_dep_data_test.bzl
index 03da9fc..9a1482c 100644
--- a/rs/private/workspace_dep_data_test.bzl
+++ b/rs/private/workspace_dep_data_test.bzl
@@ -93,6 +93,10 @@ def _workspace_configurations_preserve_local_labels_impl(ctx):
             "version": "1.0.0",
             "edition": "2021",
             "manifest_path": "/workspace/app/Cargo.toml",
+            "targets": [
+                {"name": "app-cli", "kind": ["bin"], "src_path": "/workspace/app/src/main.rs"},
+                {"name": "app_native", "kind": ["cdylib"], "src_path": "/workspace/app/src/native.rs"},
+            ],
             "dependencies": [
                 {"name": "local-helper", "rename": "renamed-helper", "kind": None, "path": "/workspace/local-helper"},
                 {"name": "local-helper", "rename": "build-helper", "kind": "build", "path": "/workspace/local-helper"},
@@ -134,6 +138,8 @@ def _workspace_configurations_preserve_local_labels_impl(ctx):
     asserts.equals(env, {}, data["dev_deps_by_platform"])
     asserts.equals(env, ["@crates//:dev-1.0.0"], all_crate_deps(data, normal_dev = True, hub_name = "crates"))
     asserts.equals(env, {"@crates//:dev-1.0.0": "dev_dep"}, crate_aliases(data, normal_dev = True, hub_name = "crates"))
+    asserts.equals(env, {"app-cli": "src/main.rs"}, data.get("binaries"))
+    asserts.equals(env, {"app_native": "src/native.rs"}, data.get("shared_libraries"))
     asserts.equals(env, "2021", data["edition"])
     asserts.equals(env, "@crates//:app_lints", data["lint_config"])
     return unittest.end(env)
```

Run `bazel test //rs/private:workspace_dep_data_tests`. At `dce9b7e`, both assertions fail: the expected maps are `{"app-cli": "src/main.rs"}` and `{"app_native": "src/native.rs"}`, but both actual values are `None`. Restoring the original target-extraction loop and the two output fields makes both tests pass. The full private suite also passes after restoration: 158 tests executed with `--nocache_test_results` and sandboxing, including our existing downstream patches.

zbarsky-openai:
`workspace_dep_data()` still assigns a package-name alias to an unrenamed local dependency at `34d5c79ecf334ae4a42036f2e539a3d6aeae76c5`. That breaks packages whose `[lib].name` differs from `[package].name`, including when the consumer uses the generated `cargo_build_script` helper.

In this two-crate example, package `helper-package` declares `[lib] name = "actual_helper"`. Cargo successfully compiles a build script calling `actual_helper::value()`, but Bazel emits the implicit alias `helper_package` and fails with:

```text
error[E0433]: cannot find module or crate `actual_helper` in this scope
 --> app/build.rs:1:57
```

The fix I tested preserves the alias from each dependency mapping when translating workspace labels. Explicit Cargo renames remain intact; `None` lets the Rust provider supply the library's crate name. The same implicit package alias should be removed for local dev dependencies. Cargo and the patched Bazel build both pass the reproducer, and all 158 `//rs/private` tests pass after updating the implicit-alias expectation; the explicit-rename test still passes.

<details>
<summary>Standalone reproducer (Linux x86-64)</summary>

```sh
mkdir crate-name-repro && cd crate-name-repro
mkdir upstream repro
curl -fsSL https://github.com/hermeticbuild/rules_rs/archive/34d5c79ecf334ae4a42036f2e539a3d6aeae76c5.tar.gz | tar xz -C upstream --strip-components=1
python3 - <<'PY_REPRO'
from pathlib import Path
# Avoid resolving rules_rs's unrelated internal tool workspaces in this fixture.
p = Path("upstream/MODULE.bazel")
p.write_text(p.read_text().replace('crate = use_extension("//rs:extensions.bzl", "crate")', 'crate = use_extension("//rs:extensions.bzl", "crate", dev_dependency = True)'))
files = {'Cargo.toml': '[workspace]\nmembers = ["app", "helper"]\nresolver = "2"\n', 'BUILD.bazel': 'exports_files(["Cargo.toml", "Cargo.lock"])\n', 'MODULE.bazel': 'module(name = "workspace_crate_names")\nbazel_dep(name = "rules_rs", version = "0.0.0")\nlocal_path_override(module_name = "rules_rs", path = "../upstream")\nbazel_dep(name = "llvm", version = "0.8.18")\nregister_toolchains("@llvm//toolchain:all")\ncrate = use_extension("@rules_rs//rs:extensions.bzl", "crate")\ncrate.from_cargo(name = "cargo", cargo_lock = "//:Cargo.lock", cargo_toml = "//:Cargo.toml", platform_triples = ["x86_64-unknown-linux-gnu"])\nuse_repo(crate, "cargo")\ntoolchains = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")\ntoolchains.toolchain(edition = "2021", version = "1.95.0")\nuse_repo(toolchains, "default_rust_toolchains")\nregister_toolchains("@default_rust_toolchains//...")\n', 'app/Cargo.toml': '[package]\nname = "app"\nversion = "0.1.0"\nedition = "2021"\n[build-dependencies]\nhelper-package = { path = "../helper" }\n', 'app/build.rs': 'fn main() { println!("cargo:rustc-env=HELPER_VALUE={}", actual_helper::value()); }\n', 'app/BUILD.bazel': 'load("@cargo//:defs.bzl", "cargo_build_script")\nload("@rules_rs//rs:rust_test.bzl", "rust_test")\ncargo_build_script(name = "build_script", srcs = ["build.rs"], crate_root = "build.rs")\nrust_test(name = "test", srcs = ["src/lib.rs"], deps = [":build_script"])\n', 'helper/Cargo.toml': '[package]\nname = "helper-package"\nversion = "0.1.0"\nedition = "2021"\n[lib]\nname = "actual_helper"\n', 'helper/BUILD.bazel': 'load("@rules_rs//rs:rust_library.bzl", "rust_library")\nrust_library(name = "helper", crate_name = "actual_helper", srcs = ["src/lib.rs"], visibility = ["//visibility:public"])\n', 'app/src/lib.rs': '#[test]\nfn value_from_build_dependency() { assert_eq!(env!("HELPER_VALUE"), "42"); }\n', 'helper/src/lib.rs': 'pub fn value() -> u32 { 42 }\n'}
for name, text in files.items():
    p = Path("repro") / name
    p.parent.mkdir(parents=True, exist_ok=True)
    p.write_text(text)
PY_REPRO
cd repro
cargo test --offline -p app
bazel test //app:test
```

The one-line `dev_dependency` adjustment only avoids resolving unrelated rules_rs tool workspaces; the workspace dependency producer is unchanged for the failing run.
</details>

The key simplification is:

```starlark
def _workspace_deps(deps, local_deps):
    return {local_deps.get(label, label): alias for label, alias in deps.items()}
```

Here `local_deps` maps each external workspace label directly to its local label, rather than also storing `dep["name"]` as an alias.

pawelchcki:
thanks this works for my edgecase too

PR #164, opened 2026-06-25
fix: cross-crate DEP_<LINKS>_INCLUDE ${out_dir} token left unresolved (#163)

### LINKED ISSUE #163: Cross-crate `DEP_<LINKS>_INCLUDE` propagation broken: `${out_dir}` token left unresolved in dependent build script (rocksdb + lz4 repro)
## Summary

When a `-sys` crate exposes an include directory via `cargo:include=$OUT_DIR/include` (cargo's `links` convention), a **dependent** crate's build script that reads `DEP_<LINKS>_INCLUDE` receives a path containing a literal, unresolved `${out_dir}` token instead of the producing crate's real `OUT_DIR`. The dependent's C/C++ compile then can't find the header.

Concretely: `librocksdb-sys`'s build script compiles RocksDB's C++ with the LZ4 backend (`-DLZ4=1`, `#include <lz4.h>`) and gets the include dir from `DEP_LZ4_INCLUDE` (set by `lz4-sys`, which has `links = "lz4"`). The build fails:

```
rocksdb/util/compression.h:49:10: fatal error: 'lz4.h' file not found
   49 | #include <lz4.h>
      |          ^~~~~~~
```

The failing `cc` invocation's include flags contain `-I <execroot>/${out_dir}/include` — the literal token `${out_dir}`, never resolved.

This is **not** a missing feature — the `links` / `DEP_<LINKS>_<KEY>` propagation is fully wired; it's a path-redaction bug that only manifests across crate boundaries.

## Root cause

The propagation chain (in the `hermeticbuild/rules_rust` fork that rules_rs provisions) is correct:

- `rs/rust_crate.bzl` passes `link_deps = deps`, so `librocksdb-sys`'s build script gets `lz4-sys` as a link-dep.
- `cargo/private/cargo_build_script.bzl` feeds each link-dep's `.depenv` file into the dependent build script.
- `cargo/private/cargo_build_script_runner/lib.rs` (~L173-204) turns `lz4-sys`'s `cargo:include=$OUT_DIR/include` into `DEP_LZ4_INCLUDE`, and `bin.rs` (~L107-128) injects the `DEP_*` vars into the dependent (rocksdb) build-script env.

The bug is in **`cargo_build_script_runner/lib.rs` `redact_paths` (~L331-341)**: it rewrites the **producing** crate's `OUT_DIR` to a literal `${out_dir}` token when writing the cross-crate `.depenv`. Its own comment states the flawed assumption:

> Safe because env files are only consumed by the target that directly owns the build script.

That assumption does not hold for `DEP_*` propagation — the `.depenv` is consumed by the **dependent's** build script, whose runner (`bin.rs` ~L114-117) only substitutes `${pwd}` and leaves `${out_dir}` literal.

Proven on disk — `lz4-sys`'s `_bs.depenv` literally contains:

```
DEP_LZ4_ROOT=${pwd}/${out_dir}
DEP_LZ4_INCLUDE=${pwd}/${out_dir}/include
```

## Minimal reproduction

`rocksdb` (with only the `lz4` feature) is the smallest real trigger. Four files + `.bazelversion`:

**`Cargo.toml`**
```toml
[package]
name = "rocksdb-repro"
version = "0.0.0"
edition = "2021"

[dependencies]
rocksdb = { version = "0.24.0", default-features = false, features = ["bindgen-runtime", "lz4"] }
```
(resolves `librocksdb-sys 0.17.3+10.4.2`, `lz4-sys 1.11.1+lz4-1.10.0`)

**`MODULE.bazel`**
```python
module(name = "rocksdb_repro")

bazel_dep(name = "rules_cc", version = "0.2.19")
bazel_dep(name = "platforms", version = "1.1.0")
bazel_dep(name = "rules_rs", version = "0.0.93")

toolchains = use_extension("@rules_rs//rs/toolchains:module_extension.bzl", "toolchains")
toolchains.toolchain(edition = "2021", version = "1.94.1")
use_repo(toolchains, "default_rust_toolchains")
register_toolchains("@default_rust_toolchains//:all")

crate.from_cargo(
    name = "crates",
    cargo_lock = "//:Cargo.lock",
    cargo_toml = "//:Cargo.toml",
    platform_triples = [
        "aarch64-apple-darwin",
        "aarch64-unknown-linux-gnu",
        "x86_64-apple-darwin",
        "x86_64-unknown-linux-gnu",
    ],
)
use_repo(crate, "crates")
```

**`BUILD.bazel`**
```python
load("@crates//:defs.bzl", "aliases", "all_crate_deps")
load("@rules_rs//rs:rust_binary.bzl", "rust_binary")

rust_binary(
    name = "repro",
    srcs = ["src/main.rs"],
    aliases = aliases(),
    edition = "2021",
    deps = all_crate_deps(normal = True),
)
```

**`src/main.rs`**
```rust
fn main() {
    let db = rocksdb::DB::open_default("/tmp/rocksdb_repro_db").unwrap();
    db.put(b"key", b"value").unwrap();
    println!("{:?}", db.get(b"key").unwrap());
}
```

**`.bazelversion`**: `9.1.0`

Then `cargo generate-lockfile && bazel build //...` → fails with `'lz4.h' file not found`.

**Environment:** rules_rs `0.0.93` (provisions `hermeticbuild/rules_rust` @ `0a4149d8a1af5f5881e73f7e2d5ff490b633fea4`), Bazel `9.1.0`, macOS arm64. The bug is platform-independent.

## Expected vs. actual

- **Expected:** rocksdb's build script receives `DEP_LZ4_INCLUDE` pointing at `lz4-sys`'s real `OUT_DIR/include`, so `#include <lz4.h>` resolves.
- **Actual:** `DEP_LZ4_INCLUDE=${pwd}/${out_dir}/include` (literal `${out_dir}`), so the compile's `-I` points at a non-existent `<execroot>/${out_dir}/include`.

## Suggested fix

Either:
- have `redact_paths` preserve the producing crate's real `OUT_DIR` (relative to execroot) for cross-crate `.depenv` entries rather than the ambiguous `${out_dir}` token, or
- have `bin.rs` resolve `${out_dir}` against each link-dep's own `out_dir` when injecting that link-dep's `DEP_*` vars into a dependent's env.

## Workaround (for reference)

Turn the `-sys` crate's build script off, link the BCR cc_library for the object code, and set `DEP_<LIB>_INCLUDE` on the dependent yourself, pointing at a materialized header directory:

```python
crate.annotation(
    crate = "lz4-sys",
    gen_build_script = "off",
    deps = ["@lz4//:lz4", "@lz4//:lz4_hc", "@lz4//:lz4_frame"],
)
crate.annotation(
    crate = "librocksdb-sys",
    build_script_data = ["@@//bazel/third_party/lz4:lz4_include"],
    build_script_env = {"DEP_LZ4_INCLUDE": "$(execpath @@//bazel/third_party/lz4:lz4_include)"},
)
```

…plus a small rule that copies the cc_library's headers into a `declare_directory` tree artifact (cc-rs's `config.include()` needs a *directory*; `$(execpath)` of a header gives a file). This works around the broken propagation, but the upstream fix above would remove the need for it entirely.

esnunes:
## Problem

A `-sys` crate that exposes an `OUT_DIR`-relative include directory via the cargo `links`/metadata convention (e.g. `cargo:include=$OUT_DIR/include`) propagated `DEP_<LINKS>_INCLUDE` to a **dependent** crate's build script with the producing crate's `OUT_DIR` rewritten to a literal, unresolved `${out_dir}` token. The dependent's C/C++ compile then pointed `-I` at `<execroot>/${out_dir}/include` and failed.

Concretely, `librocksdb-sys` (LZ4 backend) could not find `lz4.h`:

```
rocksdb/util/compression.h:49:10: fatal error: 'lz4.h' file not found
```

Fixes #163.

## Root cause

In the rules_rust build-script runner (`cargo/private/cargo_build_script_runner/lib.rs`), `outputs_to_dep_env` used `redact_paths`, which rewrites the producing crate's `OUT_DIR` to the generic `${out_dir}` token. That token is only resolved by the **rustc** action's `process_wrapper` (via `--out-dir`), for the crate that *owns* the build script. A cross-crate `.depenv` file is instead consumed by a **dependent** crate's build-script runner (`bin.rs`), which substitutes only `${pwd}` — so `${out_dir}` survived verbatim.

## Fix

`outputs_to_dep_env` now redacts only the exec root (`${pwd}`) and keeps the producing crate's real `out_dir`-relative path. That output-directory tree is already a declared input of the dependent build-script action at that same exec-root-relative path, so `${pwd}/<out_dir>/…` resolves correctly. Build-script actions don't advertise `supports-path-mapping`, so `--experimental_output_paths=strip` won't rewrite the path either.

Because rules_rust is provisioned as a pinned `http_archive` (not vendored), the fix ships as an in-repo patch applied through the existing `rules_rust.patch` extension tag (same mechanism as `rs/private/pyo3/patches/`).

## Tests (in the patch)

- **Unit:** `dep_env_out_dir_is_preserved_not_tokenized` (+ a Windows-path variant) assert the exact redacted string.
- **End-to-end:** extends `test/cargo_build_script/metadata_dep_env` — a producer build script emits `cargo:include=$OUT_DIR/include` (with a real header); the consumer reads `DEP_PRODUCER_INCLUDE` and a new `metadata_dep_env_include_test` asserts it resolves to an existing path.

## Verification

- Runner unit tests + the `metadata_dep_env` e2e suite pass.
- **Negative control:** the e2e include test fails without the `lib.rs` fix (proving it's a real regression guard).
- The exact issue reproduction (`rocksdb 0.24.0`, features `bindgen-runtime` + `lz4`) now **builds to completion** and the binary runs (writes/reads a key) — previously it failed instantly on `lz4.h`.

## Also included

A repo-root `mise.toml` (separate `chore` commit) providing a host `cargo`/`rustc` and the `bazelisk` launcher plus `test`/`test-e2e`/`gazelle` tasks. It's incidental dev tooling used to reproduce/verify this fix — happy to split it into its own PR if preferred.


dzbarsky:
Thanks for your PR! I'm about to go on a trip so I might not get to this one for a week or so - these things are a bit tricky and I want to make sure we're giving it our due diligence

PR #167, opened 2026-06-26
Add crate_name and edition to DEP_DATA and expose helpers in defs.bzl

TvdW:
Macro callers can now query a workspace crate's name and Rust edition via the generated crate_name() and edition() functions, consistent with how aliases() works.

### DIFF
--- a/rs/extensions.bzl
+++ b/rs/extensions.bzl
@@ -571,6 +571,20 @@ def aliases(package_name = None):
 
     return dep_data["aliases"]
 
+def crate_name(package_name = None):
+    dep_data = DEP_DATA.get(package_name or native.package_name())
+    if not dep_data:
+        return None
+
+    return dep_data["crate_name"]
+
+def edition(package_name = None):
+    dep_data = DEP_DATA.get(package_name or native.package_name())
+    if not dep_data:
+        return None
+
+    return dep_data["edition"]
+
 def all_crate_deps(
         normal = False,
         normal_dev = False,
--- a/rs/private/cargo_workspace_graph.bzl
+++ b/rs/private/cargo_workspace_graph.bzl
@@ -760,10 +760,12 @@ def workspace_dep_data(
             "build_deps_by_platform": build_deps_by_platform,
             "crate_features": crate_features,
             "crate_features_by_platform": crate_features_by_platform,
+            "crate_name": package["name"].replace("-", "_"),
             "deps": deps,
             "deps_by_platform": deps_by_platform,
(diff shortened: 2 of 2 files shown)


TvdW:
This makes it easier for rules_rs users to write a macro that simply takes whatever is in the Cargo.toml and builds it.

Maybe in the future we can even have a macro inside of rules_rs that does it?

dzbarsky:
> This makes it easier for rules_rs users to write a macro that simply takes whatever is in the Cargo.toml and builds it.
> 
> Maybe in the future we can even have a macro inside of rules_rs that does it?

Yes! I have a few example such macros floating around but haven't settled on exactly what it should do. It's a bit sketchy to try to replicate all cargo logic as the unit/integration tests are non trivial and people may want to customize the logic a bit/etc. but perhaps we can add one as an example as well as the factored pieces so you can easily assemble your own! I think we also need to expose features :)

I'm traveling right now but will review this PR ASAP!

TvdW:
Surprisingly _some_ features are already there if you access DEP_DATA directly. Here's my macro today, with the patch from this PR applied:

```
def _my_macro_common(
        name,
        use_rule,
        extra_compile_data = [],
        extra_deps = [],
        test_data = [],
        test_deps = [],
        test_compile_data = [],
        **kwargs):
    data = DEP_DATA[native.package_name()]
    default_test_files = native.glob(["src/**"], exclude = ["src/**/*.rs"], allow_empty = True)

    use_rule(
        name = name,
        srcs = native.glob(["src/**/*.rs"]),
        compile_data = ["Cargo.toml"] + extra_compile_data,
        edition = edition(),
        crate_name = crate_name(),
        crate_features = data["crate_features"],
        deps = all_crate_deps(normal=True) + extra_deps,
        **kwargs
    )

    rust_test(
        name = name + "_test",
        crate = ":" + name,
        crate_features = data["crate_features"],
        compile_data = default_test_files + test_compile_data,
        data = default_test_files + test_data,
        deps = all_crate_deps(normal=True, normal_dev=True) + extra_deps + test_deps,
    )

def my_rust_library(name, **kwargs):
    _my_macro_common(
        name = name,
        use_rule = rust_library,
        **kwargs,
    )

def my_rust_binary(name, **kwargs):
    _my_macro_common(
        name = name,
        use_rule = rust_binary,
        **kwargs
    )

def my_rust_proc_macro(name, **kwargs):
    _my_macro_common(
        name = name,
        use_rule = rust_proc_macro,
        **kwargs
    )
```

PR #173, opened 2026-07-08
Add support for per-platform aliases

sin-ack:
Aliases for first-party crates were unconditionally added to the shared alias data for a workspace crate.  This does not work if a first-party dependency should only be added for specific targets (e.g. a tool can make network connections using a first-party dependency but only on certain platforms, and does not use the dependency crate otherwise).

This change makes it so that we only add an alias for a first-party crate if it would also appear as a dependency for that specific target. Dependency data now has a new field called `aliases_by_platform`, and the generated aliases() helper emits a select expression if there are platform-specific aliases.

### DIFF
--- a/rs/extensions.bzl
+++ b/rs/extensions.bzl
@@ -558,6 +558,7 @@ cargo_lints(
 
     defs_bzl_contents = \
         """load(":data.bzl", "DEP_DATA")
+load("@rules_rs//rs/private:aliases.bzl", _aliases = "aliases")
 load("@rules_rs//rs/private:all_crate_deps.bzl", _all_crate_deps = "all_crate_deps")
 
 _PLATFORMS = [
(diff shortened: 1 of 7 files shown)


dzbarsky:
> party crates were unconditionally added to the shared alias data for a workspace crate. This does not work if a first-party dependency should only be added for specific targets (e.g. a tool can make network connections using a first-party dependency but only on certain platforms, and does not use the dependency crate otherwise).
> 
> This change makes it so that we only add an alias for a first-party crate if it would also appear as a dependency for that specific target. Dependency data now has a new field called `aliases_by_platform`, and the generated aliases() helper emits a select expression if there are platform-specific aliases.

So fun fact, it used to be like this but then I convinced myself that aliases were lazy enough that it was fine to merge for simplicity (i.e. that it wouldn't actually add any inputs, only remap deps if they were selected). It sounds like that might have been an incorrect assumption, but can you show me a case where it breaks? (I saw you added a unit test, but I'd prefer an integration test for this)

PR #175, opened 2026-07-08
Set CARGO_MANIFEST_PATH for generated crates

### LINKED ISSUE #174: Set CARGO_MANIFEST_PATH for generated crates
AI-made, human-reviewed.

## Summary

`rules_rs` generates Cargo package environment variables such as `CARGO_PKG_NAME`, but it does not provide `CARGO_MANIFEST_PATH` for generated crates.

Cargo documents `CARGO_MANIFEST_PATH` as available to crates and build scripts. Code using `env!("CARGO_MANIFEST_PATH")` or build scripts reading `std::env::var("CARGO_MANIFEST_PATH")` fails under Bazel with `rules_rs`.

## Reproduction

Create a path dependency crate with:

```rust
pub const CARGO_MANIFEST_PATH: &str = env!("CARGO_MANIFEST_PATH");
```

and a build script with:

```rust
let manifest_path = std::env::var("CARGO_MANIFEST_PATH")?;
assert!(std::path::Path::new(&manifest_path).is_file());
```

Then build that path dependency through `crate.from_cargo`.

## Expected Behavior

`CARGO_MANIFEST_PATH` is set for generated crate compilation and build scripts, and it points to the generated crate's `Cargo.toml`.

## Actual Behavior

Compilation fails when Rust code uses `env!("CARGO_MANIFEST_PATH")`, because the variable is not defined. Build scripts that read the variable also cannot rely on it.

## Proposed Fix

Add `CARGO_MANIFEST_PATH` to the generated `cargo_toml_env_vars.env` in `rs/private/repository_utils.bzl`.

The path should include the execution-time `${pwd}` prefix so build scripts can resolve it as a file path:

```starlark
cargo_manifest_root = "${pwd}/external/%s/%s" % (rctx.name, package_path) if package_path else "${pwd}/external/%s" % rctx.name
cargo_manifest_path = "%s/Cargo.toml" % cargo_manifest_root
```

and then:

```starlark
"CARGO_MANIFEST_PATH": cargo_manifest_path,
```

## Regression Coverage

Add a small `test/cargo_manifest_path_env` fixture with a path dependency that verifies both:

- compile-time `env!("CARGO_MANIFEST_PATH")`
- build-script access to `CARGO_MANIFEST_PATH` as an existing `Cargo.toml` file

## References

- Cargo environment variables: https://doc.rust-lang.org/cargo/reference/environment-variables.html
- Rust `env!` macro: https://doc.rust-lang.org/std/macro.env.html

TrapsterDK:
AI-made, human-reviewed.

Fixes #174.

## Summary

- add `CARGO_MANIFEST_PATH` to generated Cargo env files
- use `${pwd}/external/.../Cargo.toml` so build scripts can resolve the manifest path at execution time
- add a regression fixture that validates both `env!("CARGO_MANIFEST_PATH")` and build-script access to the manifest path as an existing file

## Testing

- `cargo fmt --manifest-path test/cargo_manifest_path_env/Cargo.toml`
- `cargo fmt --manifest-path test/cargo_manifest_path_env/manifest_path_probe/Cargo.toml`
- `bazel run --bes_backend= //.github/workflows:buildifier.check`
- `bazel build --bes_backend= --verbose_failures @cargo_manifest_path_env//:manifest_path_probe-0.1.0`


rs/private/repository_utils.bzl
```diff
@@ -85,11 +85,15 @@ def cargo_build_file_values(rctx, cargo_toml, gen_binaries, package_path = "", g
     version = package["version"]
     parsed_version = parse_full_version(version)
 
+    cargo_manifest_root = "${pwd}/external/%s/%s" % (rctx.name, package_path) if package_path else "${pwd}/external/%s" % rctx.name
```
dzbarsky: hmm I'm a little worried about this, I think it might not work correctly with sibling-repository-layout flag flipped and hardcoding "external" is a bit of an impl detail.

I wonder if we can instead do something like `$(execpath :Cargo.toml)`? I think it would need to be wired on build_script_env/rustc_env instead of going into `cargo_toml_env_vars.env`. Mind having your bot give this a shot?

TrapsterDK: You were right about this. That was a cleaner way to solve it, I just wasn't familiar enough with the code to see the better option. I have updated the code.

dzbarsky:
Thanks! I'm working through some other patches and will cut a release soon

PR #179, opened 2026-07-13
Make bindgen apple-aware

adincebic:
Use `apple_support.run(...)` when bindgen executes on macOS.


rs/rules_rust_bindgen.bzl
```diff
@@ -121,6 +137,44 @@ def _rust_bindgen_impl(ctx):
...
+        wrapper = ctx.actions.declare_file(ctx.label.name + "_apple_bindgen_wrapper.sh")
+        ctx.actions.write(
+            wrapper,
+            """#!/bin/bash
+set -euo pipefail
+
+bindgen="$1"
+shift
+
+resource_dir="$(xcrun clang -print-resource-dir)"
```
dzbarsky, line 150: I think ideally the apple_support toolchain would compute this resource-dir in the discovery repo rule and add that to the toolchain's args explicitly, instead of working around it here.

Failing that, I suppose [we can bring back this logic](https://github.com/hermeticbuild/rules_rust/blob/f19661ba71f77c993ce1eb082053692610fe4313/extensions/bindgen/private/bindgen.bzl#L192-L207) which I wasn't a fan of.

@keith wdyt?

keith: I guess it doesn't do that since we happily rely on the default? how does bindgen discover that? not from the active compiler?

dzbarsky: Yeah, right now apple_support relies on default sibling. Bindgen in rules_rust had logic above to try to guess it. In rules_rs version I removed it and expect cc toolchain to provide all flags it needs (explicit resource dir or -isystem etc)

keith: I don't think it's safe to assume many toolchains will have that? quick check shows toolchains_llvm does pass it, but rules_cc, apple_support, and musl-toolchains don't. It would be a bit annoying for us since we'd have to compute it in a repo rule that we don't otherwise do stuff like this in, and then we'd have to relativize the non-hermetic path to one of the placeholder things variables, which also means then you'd have to be handling those replacements (since otherwise it would contain the absolute path to Xcode)

dzbarsky: @keith both hermetic-llvm and toolchains_llvm have it, I think musl-toolchain is not very useful now that hermetic-llvm exists, and I'm not sure how much I care about the rules_cc auto toolchains...

But if you think computing it properly in apple_support is too annoying, I guess we can bring back the old detection method,[ it's just super janky](https://github.com/hermeticbuild/rules_rust/blob/f19661ba71f77c993ce1eb082053692610fe4313/extensions/bindgen/private/bindgen.bzl#L192-L207)...

keith: I don't think that version works for non-hermetic toolchains since all_files doesn't contain those files either?

dzbarsky: so has this never worked with rules_rust? @adincebic do you know?

adincebic: I think it did, let me check and report back.

dzbarsky: btw i was thinking, what if you configure your own bindgen toolxhain that invokes xcrun in a wrapper and then execs underlying bindgen. could even be a wrapper peocided by apple_support that you compose. feels a bit cleaner to me than having rust learn about this?

PR #180, opened 2026-07-13
Fix WASM static library extension selection

adincebic:
Removing `@platforms//os:none`, to fix WASM static libraries configured without an extension.

### DIFF
--- a/rs/toolchains/declare_rustc_toolchains.bzl
+++ b/rs/toolchains/declare_rustc_toolchains.bzl
@@ -1,3 +1,5 @@
+"""Definitions for declaring Rust compiler toolchains."""
+
 load("@rules_rust//rust:toolchain.bzl", "rust_toolchain")
 load("@rules_rust//rust/platform:triple.bzl", _parse_triple = "triple")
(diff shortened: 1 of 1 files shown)


rs/toolchains/declare_rustc_toolchains.bzl
```diff
@@ -108,15 +134,7 @@ def declare_rustc_toolchains(
                 "@platforms//os:windows": ".exe",
                 "//conditions:default": "",
             }),
-            staticlib_ext = select({
-                "@llvm//constraints/windows/abi:gnu": ".a",
-                "@llvm//constraints/windows/abi:gnullvm": ".a",
-                "@llvm//constraints/windows/abi:msvc": ".lib",
-                "@platforms//os:none": "",
```
dzbarsky, line 115: can we just remove `"@platforms//os:none": "",` line instead of doing this PR? that way `"//conditions:default": ".a",` will take effect?

adincebic: Ah yes.

dzbarsky (approved):

PR #182, opened 2026-07-18
feat: Add support for Rust's stable RV32IMAC target

Geethree:
This change:

- Adds the target to the supported matrix.
- Distinguishes RV32IMAC from RV32IMC with an ISA constraint.
- Keeps RV32IMC as the default for compatibility.
- Uses the Rust distribution's hermetic `rust-lld`.

While this is important to me for my current ESP32-C6 project, it will also benefit ESP32-H2 and other bare-metal devices using this standard Rust target.

## Validation

Validated with `rules_rs` v0.0.96 and Rust 1.96.1 by building a statically linked ESP32-C6 firmware ELF.

The related RISC-V `target_arch` normalization is proposed separately as an independent prerequisite.

### DIFF
--- a/rs/platforms/constraints/BUILD.bazel
+++ b/rs/platforms/constraints/BUILD.bazel
@@ -20,6 +20,26 @@ constraint_value(
 )
 
 constraint_setting(
+    # Bare-metal RISC-V target triples can share the same Bazel CPU and OS
+    # constraints while enabling different ISA extensions. Default to the
+    # previously supported RV32IMC target so existing custom riscv32/none
+    # platforms continue to resolve to the same Rust toolchain.
+    name = "riscv_isa",
+    default_constraint_value = "rv32imc",
+)
+
+constraint_value(
+    name = "rv32imc",
+    constraint_setting = "riscv_isa",
+)
+
+constraint_value(
+    # ESP32-C6 and ESP32-H2 use the atomic-capable RV32IMAC target.
+    name = "rv32imac",
+    constraint_setting = "riscv_isa",
+)
+
+constraint_setting(
     # Disambiguates ARM soft-float vs hard-float target triples that otherwise
     # project to identical constraint sets (e.g. arm-unknown-linux-gnueabi vs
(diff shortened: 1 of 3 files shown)


rs/platforms/constraints/BUILD.bazel
```diff
@@ -19,6 +19,26 @@ constraint_value(
     constraint_setting = "linux_libc",
 )
 
+constraint_setting(
```
dzbarsky, line 22: i think this is fine to make progress but @fmeum has been defining these systematically in platforms_contrib and may have opinions here

rs/toolchains/declare_rustc_toolchains.bzl
```diff
@@ -102,6 +102,8 @@ def declare_rustc_toolchains(
             llvm_cov = "@llvm//tools:llvm-cov",
             llvm_profdata = "@llvm//tools:llvm-profdata",
             linker = select({
+                "@rules_rs//rs/platforms/config:riscv32imac-unknown-none-elf": "{}rust-lld".format(rustc_repo_label),
```
dzbarsky: perhaps we should define `lld_label = rustc_repo_label + "rust-lld"` above and reuse that in all of these? :)

rs/platforms/constraints/BUILD.bazel
Geethree: I’d be happy to make a PR there if @fmeum would prefer. Let me know.

rs/toolchains/declare_rustc_toolchains.bzl
Geethree: That makes sense. I’ll get to this sometime this weekend.

rs/platforms/constraints/BUILD.bazel
fmeum: We don't have risc coverage yet in platforms_contrib. This looks reasonable at a glance and these targets can probably become aliases when we add support in platforms_contrib.

Geethree: @fmeum Would you like me to take a stab at risc coverage in platforms_contrib?

PR #184, opened 2026-07-20
fix: detect proc-macro crates declared via crate-type

adincebic:
A library can declare itself a proc macro with `crate-type = ["proc-macro"]` instead of `proc-macro = true`; the manifest check now handles both spellings via a shared `cargo_toml_is_proc_macro` helper.

### DIFF
--- a/rs/private/cargo_toml_utils.bzl
+++ b/rs/private/cargo_toml_utils.bzl
@@ -1,0 +1,11 @@
+"""Helpers for interpreting decoded Cargo.toml manifests."""
+
+def cargo_toml_is_proc_macro(cargo_toml_json):
+    """Whether a decoded Cargo.toml declares a procedural macro library."""
+    lib = cargo_toml_json.get("lib", {})
+    crate_types = lib.get("crate-type", lib.get("crate_type", [])) or []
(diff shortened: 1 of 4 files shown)


rs/private/cargo_toml_utils.bzl
```diff
@@ -0,0 +1,11 @@
+"""Helpers for interpreting decoded Cargo.toml manifests."""
+
+def cargo_toml_is_proc_macro(cargo_toml_json):
+    """Whether a decoded Cargo.toml declares a procedural macro library."""
+    lib = cargo_toml_json.get("lib", {})
+    crate_types = lib.get("crate-type", lib.get("crate_type", [])) or []
```
dzbarsky, line 9: Nit: can you early return if lib.get("proc_-macro") and only check crate_types after that? Should be a minor optimization :)

dzbarsky (approved):

dzbarsky (approved):

rs/private/cargo_toml_utils.bzl
adincebic: Yes.

PR #185, opened 2026-07-20
fix: harden resolver convergence and preserve dependency facts

adincebic:
Deep dependency and feature chains can exceed the resolver's previous 50-round threshold, while the convergence failure check could never execute. Resolution could also mutate feature lists originating from persisted dependency facts when injecting the implicit `default` feature.

## Changes

- raise the resolver round allowance beyond the previous 50-round threshold
- check for non-convergence after the resolution loop so the failure path is reachable
- copy dependency feature lists before injecting the implicit `default` feature

### DIFF
--- a/rs/private/cargo_workspace_graph_test.bzl
+++ b/rs/private/cargo_workspace_graph_test.bzl
@@ -1,5 +1,6 @@
 load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
-load(":cargo_workspace_graph.bzl", "cargo_toml_dependencies", "compute_package_fq_deps", "resolve_package_facts", "select_package_fq_dep", "split_lockfile_packages")
+load(":cargo_workspace_graph.bzl", "cargo_toml_dependencies", "compute_package_fq_deps", "new_feature_resolutions", "resolve_package_facts", "select_package_fq_dep", "split_lockfile_packages")
+load(":resolver.bzl", "resolve")
 
 def _select_package_fq_dep_uses_package_name_impl(ctx):
     env = unittest.begin(ctx)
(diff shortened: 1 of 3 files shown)


rs/private/resolver.bzl
```diff
@@ -180,8 +180,19 @@ _MAX_ROUNDS = 50
 def resolve(mctx, packages, feature_resolutions_by_fq_crate, cfg_attrs_by_triple, debug):
     # Do some rounds of mutual resolution; bail when no more changes
     dirty_package_indices = range(len(packages))
-    for i in range(_MAX_ROUNDS):
```
dzbarsky, line 183: can we just bump this to 150 or 200 or so? I'm worried going all the way to packages can allow bugs where we aren't making enough progress per round/etc. (I thought 50 should be enough for anyone lol!)

dzbarsky (approved):
lgtm but can we tweak that limit?

rs/private/resolver.bzl
adincebic: Sure.

adincebic: Bumped to 200.

dzbarsky: Wait that's not what the PR does 😂 mind taking a look?

rs/private/cargo_workspace_graph_test.bzl
```diff
@@ -265,12 +266,130 @@ def _resolve_package_facts_attaches_feature_resolutions_impl(ctx):
...
+    resolutions[-1].features_enabled[triple].add("forward")
+    resolve(None, packages, resolutions_by_crate, {}, False)
+
+    asserts.true(env, "forward" in resolutions[0].features_enabled[triple])
+    asserts.equals(env, ["//:chain-0"], sorted(resolutions[1].deps[triple]))
+    return unittest.end(env)
+
+resolve_handles_dependency_chains_deeper_than_previous_round_limit_test = unittest.make(_resolve_handles_dependency_chains_deeper_than_previous_round_limit_impl)
+
+def _resolve_handles_feature_chains_deeper_than_previous_round_limit_impl(ctx):
```
dzbarsky: Can you remove this test please? I don't think it buys much :)

dzbarsky (approved):

rs/private/cargo_workspace_graph_test.bzl
adincebic: Removed.

dzbarsky (approved):

PR #188, opened 2026-07-24
fix(lints): preserve Cargo workspace inheritance semantics

SDAChess:
Cargo requires packages to opt into workspace lints, but the generated configuration treated them as an implicit fallback and could not expose package and workspace policies independently.

Resolve package lints without implicit inheritance and add a separate workspace_cargo_lints target for Bazel crates that opt into the workspace policy.


zbarsky-openai (commented):
thanks, will review soon!

dzbarsky (commented):
I don't understand this PR, we probably need an e2e test or example or something to show how it wires up? `workspace_cargo_lints` doesn't appear reference anywhere. can you help me understand? (also take a look at the bot comments some of which allude to the same thing)

rs/private/lint_flags_test.bzl
```diff
@@ -63,25 +106,98 @@ def _priority_ordering_impl(ctx):
...
     result = cargo_toml_lint_flags({
+        "package": {"name": "foo"},
+        "workspace": {
+            "lints": {
+                "rust": {
+                    "unsafe_code": "warn",
+                },
+            },
+        },
         "lints": {"workspace": True},
```
dzbarsky: This test puts the member’s `[lints] workspace = true` and the workspace root’s `[workspace.lints]` in the same parsed `Cargo.toml`. That is valid for a package at the workspace root, but ordinary workspace members store those settings in two different manifests. In production, `rs/extensions.bzl` calls `cargo_toml_lint_flags(workspace_cargo_toml_json)` only on the root manifest; it never reads an individual member’s lint inheritance declaration or attaches `@<hub>//:workspace_cargo_lints` to that member’s `lint_config`. As a result, a member with `[lints] workspace = true` does not automatically receive the workspace lint policy. Could you add an integration test with a real workspace root and separate member manifests, verify both an opted-in and opted-out member, and either wire the selected policy into `lint_config` or document that `lint_config = "@<hub>//:workspace_cargo_lints"` must be configured explicitly?

-zbarskybot

rs/extensions.bzl
```diff
@@ -545,10 +546,26 @@ cargo_lints(
     rustdoc_lint_flags = [
         {rustdoc}
     ],
+)
+
+cargo_lints(
+    name = "workspace_cargo_lints",
```
dzbarsky: `workspace_cargo_lints` is created unconditionally in the same hub namespace used for unversioned dependency aliases and `extra_aliased_targets`. An existing `crate.annotation(..., extra_aliased_targets = {"workspace_cargo_lints": ...})`, or a workspace dependency with that package name, now produces two Bazel targets with the same name and makes the entire hub fail to load—even when `[workspace.lints]` is absent. Could you generate this target only when `[workspace.lints]` exists and explicitly detect or avoid collisions with generated aliases? Please add a regression test for an existing `extra_aliased_targets` entry with this name.

-zbarskybot

rs/private/lint_flags.bzl
```diff
@@ -55,20 +55,41 @@ def cargo_toml_lint_flags(cargo_toml_json):
...
+
+    if not lints:
+        return struct(
+            rustc_lint_flags = [],
+            clippy_lint_flags = [],
+            rustdoc_lint_flags = [],
+        )
 
     return struct(
         rustc_lint_flags = _format_flags(lints.get("rust", {}), ""),
```
dzbarsky: Cargo allows `unexpected_cfgs = { level = "deny", check-cfg = ["cfg(loom)"] }` in `[workspace.lints.rust]`, but `_parse_lint` and `_format_flags` retain only `level` and `priority`. This target therefore emits `--deny=unexpected_cfgs` without the corresponding `--check-cfg=cfg(loom)`, so valid `#[cfg(loom)]` can fail. The existing package parser has the same limitation, but exposing a workspace lint target makes that missing Cargo behavior immediately relevant to workspaces. Could you translate `check-cfg` into rustc `--check-cfg` flags and add a workspace-lint test, or explicitly document that these Cargo lint entries are not supported?

-zbarskybot

dzbarsky (commented):
Thanks for the contribution. I reviewed Cargo workspace inheritance, generated Bazel target names, and Cargo `check-cfg` handling. I left three inline comments on the cases that need clarification or changes.

CI has not run: https://github.com/hermeticbuild/rules_rs/actions/runs/30092453789 completed with `action_required` and no jobs, so the added tests are not yet verified by GitHub Actions.

-zbarskybot

SDAChess:
Thanks for the review. The problem this PR is trying to solve is that Cargo has two distinct lint configurations:

- `[lints]` is the lint configuration for a package.
- `[workspace.lints]` defines a workspace policy, which packages inherit only when they opt in with `[lints] workspace = true`.

`rules_rs` currently generates `@<hub>//:cargo_lints` from the workspace-level `Cargo.toml`, but it cannot represent these two configurations independently. In particular, virtual workspaces commonly contain only `[workspace.lints]`, so there is currently no generated Bazel lint configuration for that policy.

My intention with `workspace_cargo_lints` was to expose `[workspace.lints]` as a `LintsInfo` provider that first-party Bazel targets can consume explicitly:

```starlark
rust_library(
    name = "my_crate",
    lint_config = "@<hub>//:workspace_cargo_lints",
    ...
)
```

The separate target avoids treating workspace lints as an implicit fallback for package lints, which would not match Cargo’s inheritance semantics.

The review correctly identifies that the current PR does not demonstrate this wiring. The module extension reads the workspace manifest, but it does not read each member’s inheritance declaration or modify independently authored Bazel targets. Therefore, the generated workspace lint configuration must currently be attached explicitly through `lint_config`.

I plan to update the PR as follows:

1. Add an integration fixture with a real workspace root and separate member manifests.
2. Attach `@<hub>//:workspace_cargo_lints` to an opted-in Bazel target and leave another target unattached, demonstrating the behavior end to end.
3. Document that `lint_config` must be configured explicitly; `[lints] workspace = true` alone does not configure a user-authored Bazel rule.
4. Generate `workspace_cargo_lints` only when `[workspace.lints]` is present, handle collisions with generated aliases, and add the requested regression test.
5. Preserve Cargo’s `unexpected_cfgs.check-cfg` entries as Rust `--check-cfg` flags and cover that in the workspace-lint test.

SDAChess:
I pushed an integration test containing a concrete example of what I’m trying to achieve. It covers opted-in and opted-out workspace members and demonstrates the current explicit Bazel wiring through `lint_config`.

Let me know if you think this is worth adding to `rules_rs`, and whether workspace lint inheritance should remain explicit through `lint_config` or happen automatically from `[lints] workspace = true`.

SDAChess:
@dzbarsky what do you think of this pull request? Is this something you want to add to rules_rs?

dzbarsky:
> @dzbarsky what do you think of this pull request? Is this something you want to add to rules_rs?

Apologies, i've been traveling this week so a bit behind on OSS things. I think it generally makes sense, although as a follow-up we should extend this, IMO, to apply the lint config more automatically. (via an example `rust_crate` integration that we provide that should reflect everything in DEP_DATA)

Though I'm wondering - what is the intended usage pattern when some lints are specified in per-crate Cargo.toml? Shouldn't it be merged with the workspace-level one?

SDAChess:
Cargo does not currently merge these configurations: a crate either inherits `[workspace.lints]` with `workspace = true`, or uses its own `[lints.*]`. Combining both is an error and remains an open Cargo design request ([cargo#13157](https://github.com/rust-lang/cargo/issues/13157)).

The `rust_crate` integration sounds like the right end state. I can rework this PR to parse each member manifest and wire the appropriate lint configuration automatically from the get-go, rather than handling automation in a follow-up, if you’d prefer.

SDAChess:
Hello @dzbarsky, do you think I could get the green light to implement things that I mentioned above? This is kind of a blocker for our adoption of Bazel in https://github.com/NVIDIA/OpenShell

rs/private/lint_flags.bzl
```diff
@@ -55,20 +55,41 @@ def cargo_toml_lint_flags(cargo_toml_json):
         rustdoc_lint_flags, each a list of strings.
     """
     lints = cargo_toml_json.get("lints", {})
+
+    if lints.get("workspace") == True:
+        lints = cargo_toml_json.get("workspace", {}).get("lints", {})
+
     if not lints:
```
dzbarsky: i think we can dedupe this between this function and the one below? feels like one of them should probably be a wrapper of the other or something?

dzbarsky (approved):

dzbarsky:
> Hello @dzbarsky, do you think I could get the green light to implement things that I mentioned above? This is kind of a blocker for our adoption of Bazel in https://github.com/NVIDIA/OpenShell

I suppose this current PR is an improvement over status quo so we can land it, but would you mind deduping the implementation a bit and getting CI green? Feel free to ping when you want another look (dzbarsky on Bazel slack is best, but I'll keep an eye on this PR as well)

SDAChess:
Thank you for keeping an eye. I don't think I can join the Bazel slack, the invite available on Google has expired. If you can invite my email `sscatton@nvidia.com` I'd be happy to reach out on the Bazel slack instead.

SDAChess:
@dzbarsky I've updated my pull request to include automatic lint setup based on the manifest so it should "just work" now. I've also tried to reduce the duplication mentioned above and added some tests. I would be happy to get another look at this and get a CI run for it

rs/private/cargo_workspace_graph.bzl
```diff
@@ -670,7 +670,9 @@ def workspace_dep_data(
         cfg_match_cache,
         repo_root,
         workspace_package,
-        use_legacy_rules_rust_platforms):
+        use_legacy_rules_rust_platforms,
+        lint_configs = None):
```
dzbarsky: nit: `lint_config = {}` (this is safe in starlark unlike python)

rs/extensions.bzl
```diff
@@ -424,6 +444,31 @@ crate.annotation(
     repo_root = _normalize_path(cargo_metadata["workspace_root"])
     workspace_package = _label_directory(cargo_lock_path)
 
+    workspace_lints_present = "lints" in workspace_cargo_toml_json.get("workspace", {})
+    workspace_manifest_path = paths.join(repo_root, "Cargo.toml")
+    lint_configs = {}
+    package_lint_targets = []
+    for index, package in enumerate(cargo_metadata["packages"]):
+        manifest_path = _normalize_path(package["manifest_path"])
+        cargo_toml_json = workspace_cargo_toml_json if manifest_path == workspace_manifest_path else run_toml2json(mctx, package["manifest_path"])
```
dzbarsky: this might be a perf issue for us (5K first-party Cargo.tomls ;) ) we can try but I might have to make this opt-out-able or something

dzbarsky (approved):
thanks for going all the way, just address that lint and get CI green please! @isaacparker0 can you also take a look?

isaacparker-at (approved):

rs/extensions.bzl
SDAChess: Made this opt-in via `generate_lint_config = True` on `crate.from_cargo`. It defaults to `False`, so workspace member manifests aren’t parsed unless explicitly enabled.

SDAChess:
CI seems to have failed on a download. Otherwise it should be good now!

dzbarsky:
> CI seems to have failed on a download. Otherwise it should be good now!

Cool, thanks! Feel free to add yourself to the users section of the README if you want!

PR #193, opened 2026-07-28
feat: re-export rust_test_suite

pdeva:
## Summary

- re-export upstream `rust_test_suite` from the `rules_rs` Rust test façade
- add an integration smoke test with two test roots and a shared source

## Why

`rules_rust` already provides `rust_test_suite` for Cargo-style integration
tests, but `rules_rs` only exposed `rust_test`. Consumers therefore had to
either load from the generated `@rules_rust` repository directly or maintain
their own equivalent macro.

Exposing the suite alongside `rust_test` keeps consumers on the `@rules_rs`
public API.

## Validation

- `bazelisk build //rs:rust_test`
- `cd test && bazelisk test //:rust_test_suite_smoke`
- buildifier check on the changed Starlark files
- `rustfmt --check` on the new Rust test sources
- `git diff --check`

### DIFF
--- a/rs/rust_test.bzl
+++ b/rs/rust_test.bzl
@@ -1,3 +1,8 @@
-load("@rules_rust//rust:defs.bzl", _rust_test = "rust_test")
+load(
+    "@rules_rust//rust:defs.bzl",
+    _rust_test = "rust_test",
+    _rust_test_suite = "rust_test_suite",
+)
 
 rust_test = _rust_test
+rust_test_suite = _rust_test_suite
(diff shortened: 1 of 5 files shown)


dzbarsky:
@pdeva Thanks for sending this up! I've been thinking we should actually expose a [simplified/cleaned-up version of this](https://github.com/openai/codex/blob/4642370542739d5dd080b0c87a9de06a6435d3db/defs.bzl#L181-L660) `rust_crate` macro - the idea being that we should be generating not only the unit tests like `rust_test_suite`, but also the per-binary test targets, like Cargo does. As part of that I can decompose to some helper macros so the piece can be used independently, that can cover `rust_unit_test_suite` (note *unit* in the name) as you are doing here. WDYT of that approach?

pdeva:
@dzbarsky i like your approach 👍

PR #205, opened 2026-08-02
test: verify Cargo cfgs for all supported triples

Geethree:
## Summary

- generate baseline cfg values for every supported target with the registered Rust compiler
- use the generated values for exact matching, including repeated keys and bare predicates
- reject target triples outside `ALL_TARGET_TRIPLES` rather than approximating their cfg values
- add a generated-source freshness test and exhaustive positive/negative parser coverage

The generated `cfg_target_data.bzl` file is a 14.8 KB interned target-cfg table loaded by `cfg_parser.bzl` during normal rule evaluation.

## Performance

I profiled the same parser query against upstream `main` and this branch with Bazel 9.0.0:

```sh
bazel query //rs/private:cfg_parser --profile=profile.json
```

I extracted the sole `preloadOrThrow` duration from each trace with:

```sh
jq '[.traceEvents[] | select(.name == "preloadOrThrow") | .dur] | add' profile.json
```

Bazel reports trace durations in microseconds:

| Revision | Starlark preload |
| --- | ---: |
| upstream `main` | 11,587 us (11.6 ms) |
| this branch | 25,391 us (25.4 ms) |
| difference | +13,804 us (+13.8 ms) |

This is a single before/after measurement rather than a benchmark distribution. The percentage increase is large because the baseline is small; the measured absolute difference was 13.8 ms.

## Testing

- `bazel test //rs/private:cfg_parser_tests //rs/private:update_cfg_target_data_tests`
- `bazel test //...`


dzbarsky (commented):
@codex review

tools/cfg_oracle/README.md
```diff
@@ -0,0 +1,22 @@
...
+Regenerate the tables after changing the registered Rust version or the list
+of supported triples:
+
+```sh
+bazel run //rs/private:update_cfg_target_data
+bazel test //rs/private:cfg_parser_tests
+bazel test //rs/private:update_cfg_target_data_tests
+```
+
+The generated-source test reruns the oracle and detects stale data. Arbitrary
```
dzbarsky: is this just LLM being conservative or are arbitrary custom triples a real thing? (wondering if we can remove heuristics/fallbacks)

rs/private/cfg_parser_test.bzl
```diff
@@ -120,8 +131,35 @@ def _cfg_parser_smoke_test_impl(ctx):
...
 cfg_parser_smoke_test = unittest.make(_cfg_parser_smoke_test_impl)
 
+def _cfg_parser_oracle_test_impl(ctx):
+    env = unittest.begin(ctx)
+
+    asserts.equals(env, sorted(ALL_TARGET_TRIPLES), sorted(CFG_ATOM_IDS_BY_TRIPLE.keys()))
+
+    for triple in ALL_TARGET_TRIPLES:
+        expected_atoms = [CFG_ATOMS[atom_id] for atom_id in CFG_ATOM_IDS_BY_TRIPLE[triple]]
+        expected_set = {atom: True for atom in expected_atoms}
```
dzbarsky: you can use `set` now :)

rs/private/cfg_parser.bzl
```diff
@@ -207,48 +220,64 @@ def _abi_from_env(env):
...
-        return ctx["target_arch"] == "aarch64"
-
-    return False
-
-def triple_to_cfg_attrs(triple):
+def _cfg_context(triple, atoms):
+    values = {}
+    ctx = {
+        "_triple": triple,
+        "_values": values,
```
dzbarsky: why did all of these move into `_values` isntead of being directly on the ctx?

rs/private/cfg_parser.bzl
```diff
@@ -207,48 +220,64 @@ def _abi_from_env(env):
...
+    return _heuristic_cfg_atoms(triple)
 
-        # convenience booleans for bare predicates
-        "true": True,
-        "false": False,
-        "unix": fam == "unix",
-        "windows": fam == "windows",
-        "wasm": fam == "wasm",
-    }
+def triple_to_cfg_attrs(triple):
```
dzbarsky, line 180: this will run once per-cargo-closure. I wonder if its worth deduping the construciton of these contexts by lifting it higher in the graph and passing those in instead of the triples (since triples can be retrieved from them anyway?)

rs/private/cfg_parser.bzl
```diff
@@ -207,48 +220,64 @@ def _abi_from_env(env):
...
+    if arch_part == "x86_64":
+        atoms.append('target_feature="sse2"')
+    elif arch_part == "aarch64":
+        atoms.append('target_feature="neon"')
+    return atoms
+
+def cfg_atoms_for_triple(triple):
+    atom_ids = CFG_ATOM_IDS_BY_TRIPLE.get(triple)
+    if atom_ids != None:
+        return [CFG_ATOMS[atom_id] for atom_id in atom_ids]
```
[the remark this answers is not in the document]
Geethree: I think this is a valid find. My first thoughts are:
- Make the data version-aware: Generate tables keyed by rust version and have each cargo closure use the table matching its selected compiler.
- Enforce the recorded version: If the table was generated with rust 1.92, only use it when the selected compiler is 1.92. For any other version, fail clearly instead of silently calculating the wrong dependency graph.

Any thoughts here @dzbarsky?

rs/private/cfg_parser.bzl
Geethree: The extra map was intended to support repeated rustc values such as target_feature and target_has_atomic, but the nesting was unnecessary. The context now stores a set directly at each cfg key and generic equality uses membership against that set.

rs/private/cfg_parser_test.bzl
Geethree: Updated to use set(expected_atoms).

tools/cfg_oracle/README.md
Geethree: Good point. Custom Rust target specifications are real, but this resolver requires each platform_triples entry to map to a generated rules_rs platform/config label, and those only exist for ALL_TARGET_TRIPLES. The heuristic therefore implied support that was not available end to end. I removed it and now reject triples absent from the generated table.

rs/private/cfg_parser.bzl
Geethree: The contexts are currently constructed once in resolve_cargo_workspace_members for each Cargo closure, then reused by the resolver and workspace dependency rendering; target-expression results are cached as well. Lifting them further would only deduplicate across closures with the same triple set, so I left that plumbing unchanged pending evidence that cross-closure construction is material.

rs/private/cfg_parser.bzl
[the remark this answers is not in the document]
dzbarsky: tbh this makes me sad, because I don't think we want to be in the business of per-rust-version data. (What about nightlies? what about someone testing a new compiler that we haven't added data for yet? Fallback to heuristics? Is that better than just using latest? ;) )

Also, what does "have each cargo closure use the table matching its selected compiler" mean? The compiler used at build time is an arbitrary toolchain and the cargo extension has no idea which one will be selected.

Overall it feels weird to me that compiler version affects these, as they should really be immutable properties of the target platform IMO. I suspect (hope?) that it's really only the case for wasm/wasi where so much of it depends on compiler support. I wonder if we can just default to a fairly old/conservative compiler's output. Probably need to do more research, but wdyt?

rs/private/cfg_parser.bzl
dzbarsky: I'd expect a typical bazel repo to have 5-15 closures (there are some internal ones for rules_rs, though I should probably merge to a single one...). so yeah maybe not a huge deal

rs/private/cfg_parser.bzl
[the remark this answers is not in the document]
Geethree: Agreed on the research. I'll poke around this week!

theGlenn:
Sorry for joining this late. I ran into the same question while working on #213.

IMHO, per-version tables do not fit the current resolver. The Cargo extension does not know which Rust toolchain Bazel will later select.

Using an older table is not safe either -> a newer compiler can enable a cfg missing from that table.

I lean toward separating target identity (target_arch, target_os, ABI, and so on) from compiler-dependent capabilities (target_feature, atomic support).

#213 only fixes triple-derived identity and leaves compiler-dependent cfg handling unchanged.

PR #210, opened 2026-08-04
Support annotation link dependencies

finn-ball:
## Summary

- Add `link_deps` and platform-specific `link_deps_select` annotation attributes.
- Thread annotation link dependencies into generated crate metadata and `rust_library` targets.
- Preserve compatibility with synthetic rustc-source attributes that do not define the optional fields.


rs/rust_crate.bzl
```diff
@@ -206,6 +207,7 @@ def rust_crate(
         if is_proc_macro:
             (_rust_proc_macro if skip_deps_verification else rust_proc_macro)(**kwargs)
         else:
+            kwargs["link_deps"] = link_deps
```
dzbarsky, line 213: can you leave a comment explaining why `rust_proc_macro` doesn't need `link_deps`? Also it seems odd that it's silently ignored if it's specified, perhaps we should hard-fail and tell you to remove the annotation?

finn-ball: You can't hard fail. Some packages can include `rust_proc_macro`, `rust_binary` and `rust_library`. Added a comment instead.

PR #211, opened 2026-08-04
Support platform-selectable crate annotations

finn-ball:
## Summary

- Add `crate.annotation_select` for platform-specific build-script inputs and environment, crate features, and Rust compiler flags.
- Compose wildcard and exact-version selected payloads while preserving the existing exact-over-wildcard precedence for normal `crate.annotation` settings.
- Keep platform selection exclusively on `crate.annotation_select`; the normal annotation no longer exposes the corresponding `*_select` attributes.
- Preserve implicit Windows `gnullvm` annotations.
- Keep dependencies, link dependencies, and source-level settings on `crate.annotation`; they remain non-selectable.
- Migrate the smoke fixtures and add focused annotation-normalization and merging tests.

## Testing

- `bazel test //...`
- `cd test && bazel build @build_script_env_select//:bitflags-2.9.4`

### DIFF
--- a/rs/extensions.bzl
+++ b/rs/extensions.bzl
@@ -28,9 +28,6 @@ load("//rs/private:registry_utils.bzl", "CRATES_IO_REGISTRY", "registry_config_r
 load("//rs/private:repository_utils.bzl", "render_select")
 load("//rs/private:toml2json.bzl", "run_toml2json")
 
-# Bazel 8 does not define attr.label_list_dict.
-_label_list_dict = getattr(attr, "label_list_dict", attr.string_list_dict)
-
 def _spoke_repo(hub_name, name, version):
     s = "%s__%s-%s" % (hub_name, name, version)
     if "+" in s:
@@ -688,7 +685,7 @@ def _crate_impl(mctx):
             fail("`.from_cargo` is required. Please update %s" % mod.name)
 
         for cfg in mod.tags.from_cargo:
-            annotations = build_annotation_map(mod, cfg.name)
+            annotations = build_annotation_map(mod, cfg.name, cfg.platform_triples)
             annotations_by_hub_name[cfg.name] = annotations
             mctx.watch(cfg.cargo_lock)
             mctx.watch(cfg.cargo_toml)
@@ -721,7 +718,7 @@ def _crate_impl(mctx):
 
     for mod in mctx.modules:
         for cfg in mod.tags.from_cargo:
-            annotations = build_annotation_map(mod, cfg.name)
+            annotations = annotations_by_hub_name[cfg.name]
             effective_cargo_config = cargo_config_by_hub_name[cfg.name]
             use_home_cargo_credentials = cfg.use_home_cargo_credentials or global_use_home_cargo_credentials
 
@@ -772,7 +769,7 @@ def _crate_impl(mctx):
             effective_cargo_config = cargo_config_by_hub_name[cfg.name]
             cargo_credentials = cargo_credentials_by_hub_name[cfg.name]
 
-            annotations = build_annotation_map(mod, cfg.name)
+            annotations = annotations_by_hub_name[cfg.name]
 
             if cfg.debug:
                 for _ in range(25):
@@ -784,7 +781,7 @@ def _crate_impl(mctx):
     git_repos = {}
     for mod in mctx.modules:
         for cfg in mod.tags.from_cargo:
-            annotations = build_annotation_map(mod, cfg.name)
+            annotations = annotations_by_hub_name[cfg.name]
             for package in packages_by_hub_name[cfg.name]:
                 source = package.get("source", "")
                 if not source.startswith("git+"):
@@ -911,22 +908,41 @@ _from_cargo = tag_class(
     },
 )
 
+_ANNOTATION_COMMON_ATTRS = {
+    "crate": attr.string(
+        doc = "The name of the crate the annotation is applied to",
+        mandatory = True,
+    ),
+    "version": attr.string(
+        doc = "The version of the crate the annotation is applied to. Defaults to all versions.",
+        default = "*",
+    ),
+    "repositories": attr.string_list(
+        doc = "Repository names specified by crate.from_cargo(name=...). Defaults to all repositories.",
+    ),
+}
+
+_ANNOTATION_SELECTABLE_ATTRS = {
+    "build_script_data": attr.label_list(
+        doc = "Labels to add to a crate's `cargo_build_script::data` attribute.",
+    ),
+    "build_script_env": attr.string_dict(
+        doc = "Environment variables to add to a crate's `cargo_build_script::env` attribute.",
+    ),
+    "build_script_tools": attr.label_list(
+        doc = "Labels to add to a crate's `cargo_build_script::tools` attribute.",
+    ),
+    "crate_features": attr.string_list(
+        doc = "Features to add to a crate's `rust_library::crate_features` attribute.",
+    ),
+    "rustc_flags": attr.string_list(
+        doc = "Flags to add to a crate's `rust_library::rustc_flags` attribute.",
+    ),
+}
+
 _annotation = tag_class(
     doc = "A collection of extra attributes and settings for a particular crate.",
-    attrs = {
-        "crate": attr.string(
-            doc = "The name of the crate the annotation is applied to",
-            mandatory = True,
-        ),
-        "version": attr.string(
-            doc = "The version of the crate the annotation is applied to. Defaults to all versions.",
-            default = "*",
-        ),
-        "repositories": attr.string_list(
-            doc = "A list of repository names specified from `crate.from_cargo(name=...)` that this annotation is applied to. Defaults to all repositories.",
-            default = [],
-        ),
-    } | {
+    attrs = _ANNOTATION_COMMON_ATTRS | _ANNOTATION_SELECTABLE_ATTRS | {
         "additive_build_file": attr.label(
             doc = "A file containing extra contents to write to the bottom of generated BUILD files.",
         ),
@@ -936,24 +952,12 @@ _annotation = tag_class(
         # "alias_rule": attr.string(
         #     doc = "Alias rule to use instead of `native.alias()`.  Overrides [render_config](#render_config)'s 'default_alias_rule'.",
         # ),
-        "build_script_data": attr.label_list(
-            doc = "A list of labels to add to a crate's `cargo_build_script::data` attribute.",
-        ),
         # "build_script_data_glob": attr.string_list(
         #     doc = "A list of glob patterns to add to a crate's `cargo_build_script::data` attribute",
         # ),
-        "build_script_data_select": _label_list_dict(
-            doc = "Labels to add to a crate's `cargo_build_script::data` attribute, keyed by platform triplet.",
-        ),
         # "build_script_deps": attr.label_list(
         #     doc = "A list of labels to add to a crate's `cargo_build_script::deps` attribute.",
         # ),
-        "build_script_env": attr.string_dict(
-            doc = "Additional environment variables to set on a crate's `cargo_build_script::env` attribute.",
-        ),
-        "build_script_env_select": attr.string_dict(
-            doc = "Additional environment variables to set on a crate's `cargo_build_script::env` attribute. Key should be the platform triplet. Value should be a JSON encoded dictionary mapping variable names to values, for example `{\"FOO\": \"bar\"}`.",
-        ),
         "build_script_env_files": attr.label_list(
             doc = "Files containing additional environment variables for a crate's `cargo_build_script`.",
             allow_files = True,
@@ -977,12 +981,6 @@ _annotation = tag_class(
         "build_script_tags": attr.string_list(
             doc = "A list of tags to add to a crate's `cargo_build_script` target.",
         ),
-        "build_script_tools": attr.label_list(
-            doc = "A list of labels to add to a crate's `cargo_build_script::tools` attribute.",
-        ),
-        "build_script_tools_select": _label_list_dict(
-            doc = "Labels to add to a crate's `cargo_build_script::tools` attribute, keyed by platform triplet.",
-        ),
         # "compile_data": attr.label_list(
         # doc = "A list of labels to add to a crate's `rust_library::compile_data` attribute.",
         # ),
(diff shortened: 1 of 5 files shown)


dzbarsky:
@codex review

rs/extensions.bzl
```diff
@@ -327,7 +327,6 @@ crate.annotation(
             build_script_data = annotation.build_script_data,
             build_script_data_select = annotation.build_script_data_select,
             build_script_env = annotation.build_script_env,
-            build_script_env_files = annotation.build_script_env_files,
```
dzbarsky, line 330: what happened to this one?

rs/extensions.bzl
```diff
@@ -1070,10 +1050,53 @@ _annotation = tag_class(
     },
 )
 
+_annotation_select = tag_class(
+    doc = "A collection of build attributes applied to a crate for selected platform triples. Source attributes such as patches and workspace_cargo_toml belong on crate.annotation.",
+    attrs = {
+        "crate": attr.string(
```
dzbarsky: should we reuse the `attr` definitions with the other tag-class to ensure they have the same docs/etc? wdyt

test/MODULE.bazel
```diff
@@ -182,30 +182,36 @@ crate.annotation(
...
-    crate_features_select = {
-        "x86_64-unknown-linux-gnu": ["fips"],
-    },
     repositories = ["annotation_crate_features"],
-    rustc_flags_select = {
-        "x86_64-unknown-linux-gnu": ["--cfg=rules_rs_annotation_smoke"],
-    },
 )
-crate.annotation(
+crate.annotation_select(
```
dzbarsky, line 187: maybe it hsould have gone in the other PR but do you also want to add a test/example for `link_deps[_select]`?

rs/private/annotations.bzl
```diff
@@ -97,7 +99,59 @@ def annotation_for(annotations_by_crate, crate_name, version, hub_name):
...
+}
+
+_LIST_SELECT_FIELDS = [
+    "build_script_data_select",
+    "build_script_tools_select",
+    "crate_features_select",
+    "deps_select",
+    "link_deps_select",
+    "rustc_flags_select",
+    "target_compatible_with_select",
```
dzbarsky: i thought target_compatible_with was always autogenerated, what are we doing here?

rs/private/annotations.bzl
```diff
@@ -97,7 +99,59 @@ def annotation_for(annotations_by_crate, crate_name, version, hub_name):
...
+    "build_script_env": "build_script_env_select",
+    "build_script_tools": "build_script_tools_select",
+    "crate_features": "crate_features_select",
+    "deps": "deps_select",
+    "link_deps": "link_deps_select",
+    "rustc_flags": "rustc_flags_select",
+}
+
+_LIST_SELECT_FIELDS = [
+    "build_script_data_select",
```
dzbarsky: hmm its all of them except the `env` because it's a dict right? I wonder if maybe we can get their types of the saved `attr.xx` objects (if we do what I suggested below). And maybe we can even have a SELECTABLE_ATTRS const so then we can just check its keys instead of maintaining all these parallel lists?

rs/private/annotations.bzl
```diff
@@ -108,7 +162,21 @@ def build_annotation_map(mod, cfg_name):
         crate_map = annotations.setdefault(annotation.crate, {})
         if version_key in crate_map:
             fail("Duplicate crate.annotation for %s version %s in repo %s" % (annotation.crate, version_key, cfg_name))
-        crate_map[version_key] = annotation
+        crate_map[version_key] = _annotation_values(annotation)
+
+    for annotation in mod.tags.annotation_select:
+        if cfg_name not in (annotation.repositories or [cfg_name]):
```
dzbarsky: i know this is preexisting but perhaps `if annotation.repositories and cfg_name not in annotation.repositories` would be a clearer spelling, both here and above

rs/private/annotations.bzl
```diff
@@ -108,7 +162,21 @@ def build_annotation_map(mod, cfg_name):
...
+            continue
+
+        version_key = annotation.version or "*"
```
zbarsky-openai, line 231: `version = "*"` is documented as applying to all versions, but `build_annotation_map()` only merges tags with the identical `version_key`, and `annotation_for()` chooses the exact entry or `"*"` without composing them. An exact `crate.annotation` therefore suppresses a wildcard `crate.annotation_select`; reversing the scopes makes the exact select suppress wildcard settings such as `patches`, `gen_build_script`, and `workspace_cargo_toml`. Please compose exact and wildcard annotation values before constructing `_crate_annotation`.

-zbarskybot

rs/private/annotations.bzl
```diff
@@ -108,7 +162,21 @@ def build_annotation_map(mod, cfg_name):
...
+        crate_map = annotations.setdefault(annotation.crate, {})
+        values = crate_map.setdefault(version_key, structs.to_dict(_DEFAULT_CRATE_ANNOTATION))
```
zbarsky-openai: `crate.annotation_select` for `windows_aarch64_gnullvm` or `windows_x86_64_gnullvm` starts from `_DEFAULT_CRATE_ANNOTATION`. `annotation_for()` then returns that explicit annotation and never calls `_windows_gnullvm_implicit_annotation()`, so even adding one selected `rustc_flags` entry drops `gen_build_script = "off"`, `windows_import_lib`, and the generated `cc_import`. Please use the implicit annotation as the base for these crates before merging selected fields.

-zbarskybot

rs/private/annotations.bzl
```diff
@@ -1,3 +1,5 @@
+load("@bazel_skylib//lib:structs.bzl", "structs")
```
zbarsky-openai, line 1: `rs/private:BUILD.bazel`'s `bzl_library(name = "annotations")` still declares no `deps`, although this new load adds a dependency on `@bazel_skylib//lib:structs`. Please add that dependency, matching `rs/experimental/miri/private:compile`, so the declared Starlark dependency graph is complete.

-zbarskybot

zbarsky-openai (commented):
Additional adversarial pass. I did not repeat the existing findings about `build_script_env_files`, unsupported `deps`/`link_deps`, label-typed build-script inputs, or `target_compatible_with_select`.

-zbarskybot

rs/extensions.bzl
finn-ball: Applied!

rs/extensions.bzl
```diff
@@ -327,7 +327,6 @@ crate.annotation(
             build_script_data = annotation.build_script_data,
             build_script_data_select = annotation.build_script_data_select,
             build_script_env = annotation.build_script_env,
-            build_script_env_files = annotation.build_script_env_files,
             allow_build_script_to_detect_nonhermetic_paths = annotation.allow_build_script_to_detect_nonhermetic_paths,
```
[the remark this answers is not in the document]
finn-ball: Applied!

rs/private/annotations.bzl
```diff
@@ -97,7 +99,59 @@ def annotation_for(annotations_by_crate, crate_name, version, hub_name):
...
     return _DEFAULT_CRATE_ANNOTATION
 
-def build_annotation_map(mod, cfg_name):
+_SELECTABLE_ANNOTATION_FIELDS = {
+    "build_script_data": "build_script_data_select",
+    "build_script_env": "build_script_env_select",
+    "build_script_tools": "build_script_tools_select",
+    "crate_features": "crate_features_select",
+    "deps": "deps_select",
+    "link_deps": "link_deps_select",
```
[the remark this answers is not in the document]
finn-ball: Applied!

rs/extensions.bzl
```diff
@@ -1070,10 +1050,53 @@ _annotation = tag_class(
...
+        "repositories": attr.string_list(
+            doc = "Repository names specified by crate.from_cargo(name=...). Defaults to all repositories.",
+        ),
+        "triples": attr.string_list(
+            doc = "Platform triples to which the annotation applies.",
+            mandatory = True,
+        ),
+        "build_script_data": attr.string_list(
+            doc = "Labels to add to cargo_build_script.data for the selected triples.",
+        ),
```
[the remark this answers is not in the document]
finn-ball: Applied!

rs/private/annotations.bzl
finn-ball: Applied!

rs/private/annotations.bzl
finn-ball: Applied!

rs/private/annotations.bzl
finn-ball: Applied!

rs/private/annotations.bzl
finn-ball: Applied!

rs/private/annotations.bzl
finn-ball: Applied!

rs/extensions.bzl
finn-ball: oops

test/MODULE.bazel
finn-ball: Removed selectable `link_deps` as it might require some extra plumbing in a followup PR.

rs/private/annotations.bzl
finn-ball: removed

dzbarsky:
@codex review

zbarsky-openai (commented):
Re-reviewed the revised implementation, including a simplification pass. The earlier `build_script_env_files`, label anchoring, unsupported selected dependencies, Windows `gnullvm`, wildcard/select composition, and Skylib dependency findings are addressed. I did not duplicate the automated review's legacy `*_select` compatibility finding. The focused annotation tests pass locally.

-zbarskybot

rs/private/annotations.bzl
```diff
@@ -93,24 +95,140 @@ def annotation_for(annotations_by_crate, crate_name, version, hub_name):
...
+        entry = crate_map.setdefault(version_key, _annotation_entry())
+        _merge_annotation_select(entry["selects"], annotation, annotation.crate, version_key, cfg_name)
+
+    for crate_map in annotations.values():
+        wildcard_entry = crate_map.get("*")
+        wildcard_annotation = wildcard_entry["annotation"] if wildcard_entry else None
+        wildcard_selects = wildcard_entry["selects"] if wildcard_entry else {}
+        for version, entry in crate_map.items():
+            values = structs.to_dict(_DEFAULT_CRATE_ANNOTATION)
+            if wildcard_annotation:
```
zbarsky-openai: Can we avoid merging wildcard and exact `crate.annotation` values altogether? Before this PR, an exact normal annotation replaced the wildcard normal annotation; this change only needs selected fields to compose with whichever normal annotation wins. Choosing `exact annotation or wildcard annotation or implicit/default`, then overlaying wildcard and exact selects, removes `_merge_annotation_values()` and its field-specific merge policy. It also avoids an existing bug here: real exact tags contain `""` for omitted `patch_tool` and `strip_prefix`, while `_DEFAULT_CRATE_ANNOTATION` uses `None`, so an otherwise unrelated exact annotation currently erases nonempty wildcard values. The new test expectation that wildcard and exact `patches`/`link_deps` concatenate is also an unrelated change to the old precedence semantics.

-zbarskybot

rs/private/annotations.bzl
```diff
@@ -93,24 +95,140 @@ def annotation_for(annotations_by_crate, crate_name, version, hub_name):
...
+        for triple in platform_triples:
+            platform_values.setdefault(triple, [])
+        values[field] = platform_values
+
+def _merge_select_maps(base, override):
+    for field in _SELECT_MAP_FIELDS:
+        if field not in override:
+            continue
+        selected = dict(base.get(field, {}))
+        selected.update(override[field])
```
zbarsky-openai: This only merges the outer select map, so an exact-version select replaces the wildcard payload when both target the same field and triple. For example, wildcard Linux `rustc_flags = ["--cfg=common"]` plus exact-version Linux `rustc_flags = ["--cfg=v1"]` produces only the exact flag; `build_script_env` similarly loses unrelated wildcard variables. Please compose list payloads and merge environment keys. Keeping environment values as dictionaries until the repository-rule boundary, rather than JSON-encoding them in `_merge_annotation_select()`, would give every selected field one straightforward merge representation.

-zbarskybot

rs/extensions.bzl
```diff
@@ -721,7 +718,7 @@ def _crate_impl(mctx):
 
     for mod in mctx.modules:
         for cfg in mod.tags.from_cargo:
-            annotations = build_annotation_map(mod, cfg.name)
+            annotations = build_annotation_map(mod, cfg.name, cfg.platform_triples)
```
zbarsky-openai: `annotations_by_hub_name[cfg.name]` already stores this exact `build_annotation_map()` result above. Can the three later phases reuse that value instead of rebuilding and revalidating all tags here and again at lines 772 and 784? `build_annotation_map()` is now a multi-phase normalization pass, so computing it once is simpler, cheaper, and guarantees every consumer sees the same materialized annotations.

-zbarskybot

rs/private/annotations_test.bzl
```diff
@@ -0,0 +1,147 @@
...
+
+def _mod(annotations = [], annotation_selects = []):
+    return struct(
+        tags = struct(
+            annotation = annotations,
+            annotation_select = annotation_selects,
+        ),
+    )
+
+def _annotation(crate, version = "*", **kwargs):
```
zbarsky-openai, line 51: This helper creates sparse structs, but real `tag_class` values contain every declared attribute at its default. The production merge infers whether an attribute was omitted from those values, so these tests exercise a different and easier representation; that is why they miss the `patch_tool`/`strip_prefix` reset above. Prefer normalizing real tags to sparse overlays before merging, or make this factory reproduce the real tag defaults.

-zbarskybot

rs/extensions.bzl
```diff
@@ -911,22 +908,41 @@ _from_cargo = tag_class(
...
-        "version": attr.string(
-            doc = "The version of the crate the annotation is applied to. Defaults to all versions.",
-            default = "*",
-        ),
-        "repositories": attr.string_list(
-            doc = "A list of repository names specified from `crate.from_cargo(name=...)` that this annotation is applied to. Defaults to all repositories.",
-            default = [],
-        ),
-    } | {
+    attrs = _ANNOTATION_COMMON_ATTRS | _ANNOTATION_SELECTABLE_ATTRS | {
```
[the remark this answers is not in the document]
finn-ball: Sure, can restore that I guess. I removed it as it seemed redundant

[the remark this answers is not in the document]
finn-ball: After a short chat, we decided to keep them removed as they are redundant and it is a trivial migration

PR #212, opened 2026-08-05
Add an experimental cargo-mutants aspect and test rule

robot-head:
## What

`cargo-mutants` is cargo-only — it enumerates mutants, then copies the tree and drives `cargo build`/`cargo test` itself ([sourcefrog/cargo-mutants#77](https://github.com/sourcefrog/cargo-mutants/issues/77) is still open). This splits it at its own seam: cargo-mutants does what only it can do (syn-parse and list mutants), Bazel does the build and test.

```bzl
cargo_mutants_test(
    name = "mylib_mutants",
    test = ":mylib_test",
)
```

A normal `bazel test` target that reports caught/missed/unviable mutants. The same aspect sweeps over `//...` with no BUILD edits:

```
bazel build //... \
  --aspects=@rules_rs//rs/experimental/mutants:cargo_mutants_test.bzl%cargo_mutants_aspect \
  --output_groups=cargo_mutants
```

## How

The aspect records the crate's `rustc --test` command line in `short_path` form; the runner replays it once per mutant from the runfiles tree, patching the mutated span into a scratch copy of the sources. Only the target's own `#[cfg(test)]` tests run against each mutant — separate integration-test crates would need the whole downstream link chain replayed, not one command line.

Sweeps are slow (every mutant is a fresh link plus a test run), so both fan-out axes are wired:

- `jobs = N` — N mutants built and tested concurrently on one machine. Each job gets its own scratch source tree, output binary, **and `--out-dir`**: rustc drops intermediate `.rcgu.o` codegen-unit objects there and deletes them after linking, so jobs sharing one clobber each other mid-link.
- `shard_count = N` — Bazel's native sharding, so shards become independent test actions that spread across executors and RBE. The runner partitions with `.skip(index).step_by(total)` and touches `TEST_SHARD_STATUS_FILE`.

They compose.

## Tool provisioning

The binary is user-supplied through a `label_flag`, matching the existing `wasm-bindgen-cli` / `protoc-gen-prost` pattern:

```bzl
crate.annotation(crate = "cargo-mutants", gen_binaries = ["cargo-mutants"])
```
```
build --@rules_rs//rs/experimental/mutants:cargo_mutants_binary=@crates//:cargo-mutants__cargo-mutants
```

Vendoring was rejected: upstream ships **x86_64-only** release assets, so the `ubuntu-24.04-arm` CI leg and arm64 Macs could not download it. Unconfigured targets fail with a pointer to the setup, not a confusing missing-binary error.

This lands under `rs/experimental/` because cargo-mutants' `--list --json` schema is the integration contract and carries no stability guarantee we control.

## Also in here

`test/MODULE.bazel` restates the macOS SDK framework defaults. `@llvm//extensions:osx.bzl` does `if not frameworks: frameworks = _DEFAULT_FRAMEWORKS` with tags accumulating across *all* modules, so `3rd_party/coreaudio-sys/include.MODULE.bazel` was already suppressing the entire default set — `Security` was missing for anything reaching `ring`. Pre-existing; surfaced by this branch.

## Verification

| check | result |
|---|---|
| `cd test && bazel test //mutants/...` (`jobs=2`, `shard_count=2`) | 4/4 pass — `shard_1_of_2: 3 mutants: 3 caught, 0 missed, 0 unviable`, `shard_2_of_2: 2 mutants: 2 caught, 0 missed, 0 unviable` |
| sequential (`jobs=1`, unsharded) control | `5 mutants: 5 caught, 0 missed, 0 unviable` — exact agreement, so concurrency isn't losing or corrupting mutants |
| `under_tested_mutants_failure_test` (negative case) | still greps `MISSED` correctly under concurrency; 7.9s → 4.6s |
| root `bazel test //...` | 40/40 pass (includes buildifier) |
| `cd test && bazel test //...` | 19/20 — only `//:custom_registry_test`, which asserts `bazel` is on `PATH` inside the sandbox; fails on this macOS checkout independently of this branch |

## Deliberately not here

- Integration-test crates (`tests/*.rs` as separate targets) — needs the downstream link chain replayed. Add when someone asks.
- Vendoring cargo-mutants — see above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dzbarsky:
@robot-head hi, this looks like a cool capability! It's not clear to me from the description or PR if it needs to actually live in this repo (vs for example be a standalone module that we reference as an example). WDYT?

robot-head:
Yeah! That makes sense, let me ship it as a standalone repo and y'all can link to it if you want :) 

You should give it a try tho, will make your unit tests a lot better

robot-head:
Shipped it standalone: **https://github.com/robot-head/rules_rs_mutants**

```bzl
bazel_dep(name = "rules_rs_mutants", version = "0.0.1")
```
```bzl
load("@rules_rs_mutants//mutants:cargo_mutants_test.bzl", "cargo_mutants_test")

cargo_mutants_test(
    name = "mylib_mutants",
    test = ":mylib_test",
)
```

Same aspect and runner as this PR, just repackaged. Consumers get `rules_rs`, `hermetic_launcher`, `bazel_skylib`, `platforms` and one `serde_json` crate repo for the runner — cargo-mutants itself is a `dev_dependency` there, since you bring your own binary through the `label_flag` either way. CI is green on `bazel test //...`, and `example/` is a working setup if you want to try it.

Keeping it out of this repo is the right call for the reason you would expect in hindsight: the `cargo-mutants --list --json` schema is the integration contract, and neither of us controls its stability. It can churn on its own release cadence over there.

One thing worth pulling across independently of any of this — `test/MODULE.bazel` is currently missing the macOS SDK `Security` framework. `@llvm//extensions:osx.bzl` does `if not frameworks: frameworks = _DEFAULT_FRAMEWORKS`, and tags accumulate across *all* modules, so `3rd_party/coreaudio-sys/include.MODULE.bazel` declaring one framework suppresses the entire default set. Anything reaching `ring` is affected. Pre-existing, this branch just surfaced it.

Closing.

PR #213, opened 2026-08-07
fix: normalize cfg attributes for three-part Rust triples

theGlenn:
Hi @dzbarsky and team! Thanks again for [reaching out to us](https://github.com/xybrid-ai/xybrid/issues/423) and inviting us to upstream our Rust and LLVM patches. This is the first one.

## Summary

- Parse non-positional three-part triples, including Android EABI and bare-metal Arm targets.
- Extend architecture normalization to Arm/Thumb, x86, PowerPC, and BPF spellings.
- Derive pointer width from the normalized architecture while retaining the raw spelling for endianness.
- Match longer ABI names first so `eabihf` is not reduced to `eabi`.

`triple_to_cfg_attrs` currently interprets `armv7-linux-androideabi` as `target_os = "androideabi"` and `target_arch = "armv7"`, with no Unix family or `eabi` ABI. Dependencies gated by `cfg(unix)`, `cfg(target_os = "android")`, or `cfg(target_arch = "arm")` are therefore omitted from 32-bit Android builds.

The repository's `rules_rust` triple parser already maps `androideabi` to `android`. This change brings `cfg_parser.bzl` in line with that behavior and with `rustc --print cfg`.

## Validation

- Added focused coverage for Android EABI, bare-metal Arm, architecture aliases, pointer width, endianness, ABI precedence, and unchanged four-part triples.
- Expected values come from `rustc --print cfg --target <triple>`.
- `bazel test //rs/private:cfg_parser_tests` passes.
- `bazel test //...` passes (40/40 tests).

### DIFF
--- a/rs/private/cfg_parser.bzl
+++ b/rs/private/cfg_parser.bzl
@@ -150,14 +150,75 @@ def _normalize_os(os_raw):
         return "macos"
     return os_raw
 
+# Architecture tokens that rustc reports under a different `target_arch` than
+# the triple spells. Mirrors the `arch` field of rustc's target specs.
 def _normalize_arch(arch_raw):
+    # Apple spells 64-bit Arm arm64/arm64e/arm64_32. Preserve other arm64*
+    # architectures such as Windows ARM64EC; every remaining arm* and thumb*
+    # variant is 32-bit Arm.
+    if arch_raw in ["arm64", "arm64_32", "arm64e"]:
+        return "aarch64"
+    if arch_raw.startswith("arm64"):
+        return arch_raw
+    if arch_raw.startswith("arm") or arch_raw.startswith("thumb"):
+        return "arm"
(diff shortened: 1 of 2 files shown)


dzbarsky:
Thanks for sending this up! It looks like a similar open PR but we had an unresolved question there https://github.com/hermeticbuild/rules_rs/pull/205#discussion_r3700247105 - wonder if you have thoughts on that bit? fyi @Geethree

theGlenn:
Hey @dzbarsky. 

I finally had time to look at #205 and added my thoughts to the existing [discussion](https://github.com/hermeticbuild/rules_rs/pull/205#issuecomment-5442342757).

The PRs overlap on target identity, but they take different approaches. 
- #205 replaces the parser with a generated cfg oracle. 
- #213 is a narrow fix to the existing triple parser and leaves compiler-dependent cfg handling unchanged.

Since #205’s versioning question is still unresolved, I think #213 can land independently.

PR #217, opened 2026-08-13
feat: add Cargo-style rust_crate and reusable test suites

dzbarsky:
## Summary

Add a public `@rules_rs//rs:rust_crate.bzl` macro that infers Cargo libraries, binaries, build scripts, library and binary unit tests, and integration tests. Expose independently reusable `rust_unit_test_suite` and `rust_integration_test_suite` helpers, and re-export the upstream-compatible `rust_test_suite` requested in #193.

The existing generated-dependency `rust_crate` implementation was moved to `rs/private/rust_crate.bzl` in #218.

The new Cargo workspace fixture covers binary-only packages, default and hyphenated binaries, nested binary and integration-test roots, shared test modules, normal/development/build dependency separation and overlap, build-script helper modules, `CARGO_PKG_NAME`, `CARGO_PKG_VERSION`, and compile-time/runtime `CARGO_BIN_EXE_*`.

## Validation

- `bazel test --config=remote //rust_crate/...` in `test/`: 15 tests passed.
- Root-module public/private targets and `//rs/private:all`: 42 tests passed.
- `buildifier -mode=check`, `rustfmt --check`, and `git diff --check` passed.

Full-suite validation on this macOS checkout encounters pre-existing local conditions: an ignored `test/target` directory is interpreted as a missing `@target` repository, and an unrestricted root remote run selects a macOS executable on Linux workers. Focused and existing private regression suites pass.


rs/rust_test.bzl
```diff
@@ -1,3 +1,173 @@
...
+    for binary in binaries:
+        binary_name = _label_name(binary)
+        if binary_name.endswith("__bin"):
+            binary_name = binary_name[:-len("__bin")]
+        cargo_binary_env["CARGO_BIN_EXE_" + binary_name] = "$(rootpath %s)" % binary
+
+    caller_rustc_env = kwargs.pop("rustc_env", {})
+    caller_env = kwargs.pop("env", {})
+
+    rustc_env = dict(cargo_binary_env)
```
dzbarsky: just do `rustc_env = cargo_binary_env | rustc_env` and add it to kwargs with default of {}? Same with the other ones

README.md
```diff
@@ -128,6 +128,64 @@ rust_binary(
...
+the standard Cargo source layout:
+
+```bzl
+load("@crates//:defs.bzl", "aliases", "all_crate_deps")
+load("@rules_rs//rs:rust_crate.bzl", "rust_crate")
+
+rust_crate(
+    name = "my_crate",
+    aliases = aliases(),
+    deps = all_crate_deps(normal = True),
```
dzbarsky: drop `normal = True`, we don't need that! actually can this just take a `dep_data` struct like DEP_DATA creates?

dzbarsky (commented):
comments

PR #255, opened 2026-09-08
feat: coalesce crates into one linked lib where possible across hubs/features/versions

### LINKED ISSUE #144: Allow a bzlmod module to consume a `@crates` hub declared by another module

genevieve-me:
Resolves #144. Depends on https://github.com/hermeticbuild/rules_rust/pull/57. Hefty MR so I tried to break it into individually reviewable commits: a lot of the lines changed number is churn in Bazel lockfiles.

This change allows one Bzlmod module to share a `@crates` hub with another module without producing duplicate Rust libraries for the same Cargo package. When the same compatible Cargo package appears in multiple hubs, `rules_rs` now reuses a single generated library target. If two configured instances are not safe to link into the same binary, the build fails during analysis instead of quietly linking both copies.

Each hub records the compilation properties that affect whether two crate instances can safely be shared, and those records are combined before assigning compatible instances to a canonical generated target. If the information needed to establish compatibility is missing, that combination is rejected.

I tested this to cover compatibility edge cases like optional-dependency and feature interactions that can change the effective crate graph, e.g., `dep?/feature` interactions and dependencies whose presence or configuration depends on enabled features.

The MR split up metadata and archive handling so that crates with the same name and version from different registries can coexist correctly - I figured some users might have an internal registry with a foobar 1.0.0 that collides with crates.io foobar 1.0.0.

This will need a rules_rust dependency bump once identity is in that ruleset.

#### Behavior notes

Dependency processing now uses deterministic topological ordering, so coalescing gives the same result regardless of input ordering. Repository staging and generated names are deterministic, with explicit collision detection.

This does not introduce global Cargo dependency resolution. Each hub still owns its own lockfile and dependency closure; coalescing only determines when compatible crate instances from those independent hubs may safely share a generated library target.

Package identity from rules_rust includes the package source as well as its name and version. In the future, we could use full Cargo Package IDs instead of name and version, since right now crates with the same name and version but different sources are rejected.

The current compatibility-class construction is intentionally conservative and fails closed when link safety cannot be established.


zbarsky-openai:
This is really interesting, thanks for tackling it. I need to spend some proper time to dig in and understand what's happening here - it probably won't happen before Bazelcon. But just wanted to ack your PR and mention it's on my radar!

